PR #158, opened 2021-07-06
Fixup shadow errors and warnings

mrieggeramzn:
Fix for missing shadow map inputs in the forward pass of the msaa rpi sample

Also did some cleanup. Instead of calling multiple individual light culling components, call the LightCullingParentTemplate instead.

ATOM-15928


mrieggeramzn:
Closing because Doug fixed this in dev

PR #161, opened 2021-07-07
 [ATOM-15904] Run automated performance benchmarks in periodic test suite.

synicalsyntax:
Merging of this PR is blocked until after o3de/o3de#1923 gets merged, but reviews are still appreciated.


Standalone/PythonTests/Automated/benchmark_data_aggregator.py
```diff
@@ -0,0 +1,117 @@
+#!/usr/bin/env python
```
amzn-tommy, line 1: This is just moving the changes that were in https://github.com/o3de/o3de/pull/1845/commits/74b995f7e7c386bcca059173aa64acbd6e8398f9, correct?

synicalsyntax: Yes, but note that I modified the script-like execution into a class-based one instead.

Standalone/PythonTests/Automated/benchmark_data_aggregator.py
```diff
@@ -0,0 +1,117 @@
...
+import json
+from pathlib import Path
+import time
+import copy
+import subprocess
+import requests
+
+from ly_test_tools.mars.filebeat_client import FilebeatClient
+
+class BenchmarkDataAggregator:
```
jromnoa, line 18: nitpick: can you make this: `class BenchmarkDataAggregator(object):` ? that helps enforce the use of Python 3 with this file

Standalone/PythonTests/Automated/benchmark_data_aggregator.py
```diff
@@ -0,0 +1,117 @@
...
+    def __init__(self, workspace, logger, test_suite):
+        self.build_dir = Path(workspace.paths.build_directory())
+        self.results_dir = Path(workspace.paths.project()) / 'user' / 'Scripts' / 'PerformanceBenchmarks'
```
jromnoa, line 21: here and throughout is using `Path()` necessary? wouldn't this work just the same (and match our existing patterns for paths):
`self.results_dir = os.path.join(workspace.paths.project(), 'user', 'Scripts', 'PerformanceBenchmarks')`

Standalone/PythonTests/Automated/benchmark_data_aggregator.py
```diff
@@ -0,0 +1,117 @@
...
+        return payloads
+
+    def upload_metrics(self):
+        start_timestamp = time.time()
+
+        git_commit_data = subprocess.check_output(['git', 'rev-parse', '--short', 'HEAD'], cwd=self.build_dir)
+        git_commit_hash = git_commit_data.decode('ascii').strip()
+        benchmark_metadata = { 'gitCommit': git_commit_hash }
+
+        for benchmark_dir in self.results_dir.iterdir():
```
jromnoa, line 104: should we do an exists check on this path first? i.e. something similar to:
```
class BenchmarkPathException(Exception):
    """Custom Exception class for invalid benchmark file paths."""
    pass

if not os.path.exists(self.results_dir):
    raise BenchmarkPathException(f'results_dir does not exist at location: {self.results_dir}')
```

Standalone/PythonTests/Automated/benchmark_data_aggregator.py
```diff
@@ -0,0 +1,117 @@
...
+        self.test_suite = test_suite
+        self.logger = logger
+        self.filebeat_client = FilebeatClient(logger)
+
+    def _process_benchmark(self, benchmark_dir):
+        benchmark_metadata = {}
+        frame_stats = { 'count': 0, 'totalTime': 0 }
+        pass_stats = {}
+
+        metadata_file = benchmark_dir / 'benchmark_metadata.json'
```
jromnoa, line 31: can we use `os.path.joins()` here instead? this would match our common patterns in most of our python3 code (we don't use `/` any more)

Standalone/PythonTests/Automated/benchmark_data_aggregator.py
```diff
@@ -0,0 +1,117 @@
...
+
+class BenchmarkDataAggregator:
+    def __init__(self, workspace, logger, test_suite):
+        self.build_dir = Path(workspace.paths.build_directory())
+        self.results_dir = Path(workspace.paths.project()) / 'user' / 'Scripts' / 'PerformanceBenchmarks'
+        self.test_suite = test_suite
+        self.logger = logger
+        self.filebeat_client = FilebeatClient(logger)
+
+    def _process_benchmark(self, benchmark_dir):
```
jromnoa, line 26: here and throughout: can we get a brief docstring describing what each function does as well as what the params are expected to be? i.e. something similar to:
```
def _process_benchmark(self, benchmark_dir):
    """
    Processes benchmarks using benchmark_dir to find metadata json files that allow it to...etc...etc..
    :param benchmark_dir: path to the metadata json files and benchmark files
    :return: tuple with 3 elements: benchmark_metadata, frame_stats, pass_stats 
    """
```

Standalone/PythonTests/Automated/benchmark_data_aggregator.py
```diff
@@ -0,0 +1,117 @@
...
+        benchmark_metadata = {}
+        frame_stats = { 'count': 0, 'totalTime': 0 }
+        pass_stats = {}
+
+        metadata_file = benchmark_dir / 'benchmark_metadata.json'
+        if metadata_file.exists():
+            data = json.loads(metadata_file.read_text())
+            benchmark_metadata = data['ClassData']
+        else:
+            self.logger.warn(f'Metadata file could not be found at {metadata_file}')
```
jromnoa, line 36: is this `metadata_file` not required? if so this is fine

however, if its required we should raise an Exception or error and block progression, just wanted to call this out just in case

Standalone/PythonTests/Automated/benchmark_data_aggregator.py
```diff
@@ -0,0 +1,117 @@
...
+
+        for file in benchmark_dir.iterdir():
+            if file.is_dir() or not is_timestamp_file(file):
+                continue
+
+            data = json.loads(file.read_text())
+            entries = data['ClassData']['timestampEntries']
+
+            frame_time = 0
+            for entry in entries:
```
jromnoa, line 49: since this is a `for` loop inside of a `for` loop, its probably better we split into its own function that returns `time_ns` for updating `frame_time`
in general we try to avoid nested `for` loops and split them into new functions with returns instead (makes it easier to follow the code)

also, if we could declare the variables related to that code closer to the code, its easier to follow
i.e. `pass_stats ` is on line 29 and i had to find it to see what this `pass_entry = pass_stats.get(name, { 'totalTime': 0, 'maxTime': 0 })` call was doing

speaking of which, isn't `pass_stats` always `{}` before this call is made? that makes the `.get()` call redundant since it won't get anything matching `entry['passName']` inside `pass_stats = {}`

you should just do this and remove the `.get()` call if `pass_stats` is always `{}` before that call is made:
`pass_stats = {'totalTime': 0, 'maxTime': 0}`

jromnoa: since its iterating and looping over values it seems the `pass_entry = pass_stats.get(name, { 'totalTime': 0, 'maxTime': 0 })` may be needed, but i would initialize `pass_entry = {'totalTime': 0, 'maxTime': 0}` before this for loop / future function instead of `pass_entry = {}`

Standalone/PythonTests/Automated/benchmark_data_aggregator.py
```diff
@@ -0,0 +1,117 @@
...
+            frame_stats['maxTime'] = max(frame_time, frame_stats.get('maxTime', 0))
+            frame_stats['minTime'] = min(frame_time, frame_stats.get('minTime', frame_time))
+            frame_stats['count'] += 1
+
+        if frame_stats['count'] < 1:
+            self.logger.warn(f'No frame timestamp logs were found in {benchmark_dir}')
+
+        return benchmark_metadata, frame_stats, pass_stats
+
+    def _generate_payloads(self, benchmark_metadata, frame_stats, pass_stats):
```
jromnoa, line 70: same here for docstrings, it gets hard to follow without knowing specifically the data type and value of each `param`

Standalone/PythonTests/Automated/benchmark_runner_periodic_suite.py
```diff
@@ -0,0 +1,55 @@
...
+        # Execute test.
+        process_utils.safe_check_call(cmd, stderr=subprocess.STDOUT, encoding='UTF-8', shell=True)
+        try:
+            expected_lines = ["Script: Capturing complete."]
+            atomsampleviewer_log_monitor.monitor_log_for_lines(expected_lines)
+
+            aggregator = BenchmarkDataAggregator(workspace, logger, 'periodic')
+            aggregator.upload_metrics()
+        except ly_test_tools.log.log_monitor.LogMonitorException as e:
+            raise AtomSampleViewerException('Data capturing did not complete in time.')
```
jromnoa, line 55: can we pipe the `e` variable in here as well?
```
raise AtomSampleViewerException(f'Data capturing did not complete in time, got error: {e}')
```

jromnoa: also since there's 2 RHI values that run, you should pipe in which RHI failed, since the same runs both RHIs (dx12 and vulkan)

Standalone/PythonTests/Automated/benchmark_runner_periodic_suite.py
```diff
@@ -0,0 +1,55 @@
...
+                           '--exitontestend')
+
+        def teardown():
+            process_utils.kill_processes_named(['AssetProcessor', 'AtomSampleViewerStandalone'], ignore_extensions=True)
+        request.addfinalizer(teardown)
+
+        # Execute test.
+        process_utils.safe_check_call(cmd, stderr=subprocess.STDOUT, encoding='UTF-8', shell=True)
+        try:
+            expected_lines = ["Script: Capturing complete."]
```
jromnoa, line 49: can we add more expected lines? it looks like `Print('Capturing complete.')` is always called in the Lua script regardless of pass or fail

can we get some sort of conditional that prints `Print('Capturing complete.')` inside the Lua script? basically scared of getting false positives with the metrics, which is looks like could happen right now (correct me if im wrong)

jromnoa (changes requested):
Please also post a successful run on the PR showing pass output

Standalone/PythonTests/Automated/benchmark_data_aggregator.py
```diff
@@ -0,0 +1,143 @@
...
+        else:
+            raise BenchmarkPathException(f'Metadata file could not be found at {metadata_file}')
+
+        # data structures aggregating statistics from timestamp logs
+        frame_stats = { 'count': 0, 'totalTime': 0, 'maxTime': 0, 'minTime': float('inf') }
+        pass_stats = {}  # key: pass name, value: dict with totalTime and maxTime keys
+
+        # this allows us to add additional data if necessary, e.g. frame_test_timestamps.json
+        is_timestamp_file = lambda file: file.name.startswith('frame') and file.name.endswith('_timestamps.json')
+        def update_pass(entry):
```
jromnoa, line 50: instead of nesting this function inside `_process_benchmark()`, can we just pass in the values needed from the nesting in as params? i.e. `def update_pass(entry, pass_stats):`

that way we won't need to nest it inside `_process_benchmark()` and it can exist outside of it and also make it callable by other functions should that ever be desired in a future update

Standalone/PythonTests/Automated/benchmark_runner_periodic_suite.py
```diff
@@ -0,0 +1,56 @@
...
+    pass
+
+
+@pytest.mark.parametrize('launcher_platform', ['windows'])
+@pytest.mark.parametrize("project", ["AtomSampleViewer"])
+@pytest.mark.parametrize('rhi', ['dx12', 'vulkan'])
+@pytest.mark.usefixtures("clean_atomsampleviewer_logs", "atomsampleviewer_log_monitor")
+class TestPerformanceBenchmarksPeriodicSuite:
+    def test_BerformanceBenchmarkPeriodicSuite(self, request, workspace, launcher_platform, rhi, atomsampleviewer_log_monitor):
+        print("FDKSJHFDSHFSDKF")
```
jromnoa, line 30: what is this for? can we remove it?

jromnoa (changes requested):

Standalone/PythonTests/Automated/benchmark_data_aggregator.py
jromnoa: UPDATE: This is solved now, won't let me resolve the comment though

Standalone/PythonTests/Automated/benchmark_data_aggregator.py
jromnoa: UPDATE: This is better but let's update it to not be nested (added more info in a new review comment)

Standalone/PythonTests/Automated/benchmark_data_aggregator.py
jromnoa: UPDATE: I can't resolve this but it's resolved now

Standalone/PythonTests/Automated/benchmark_runner_periodic_suite.py
synicalsyntax: Oops, debugging statement I forgot to remove.

jromnoa (approved):

jromnoa (approved):

hershey5045 (approved):
Looks great! As Joseph mentioned, please post the output of running the test first before you merge.

synicalsyntax:
@jromnoa @hershey5045 Test output from running locally:
```
Starting 'periodic' suite: Tests which can take a long time and should be done periodially instead of every commit - these should not block code submission
Executing CTest 1 time(s) with command:
  C:\Program Files (x86)\Microsoft Visual Studio\2019\Professional\Common7\IDE\CommonExtensions\Microsoft\CMake\CMake\bin\ctest.exe --build-config profile --output-on-failure --parallel 16 --no-tests=error --label-exclude ^(SUITE_smoke|SUITE_main|SUITE_benchmark|SUITE_sandbox)$ -T Test --tests-regex AtomSampleViewer::PeriodicPerformanceBenchmarks
in working directory:
  build

Executing CTest iteration 1/1
Cannot find file: C:/Users/cyntlin/Documents/o3de/build/DartConfiguration.tcl
   Site: 
   Build name: (empty)
Cannot find file: C:/Users/cyntlin/Documents/o3de/build/DartConfiguration.tcl
Test project C:/Users/cyntlin/Documents/o3de/build
    Start 50: AtomSampleViewer::PeriodicPerformanceBenchmarks.periodic::TEST_RUN
1/1 Test #50: AtomSampleViewer::PeriodicPerformanceBenchmarks.periodic::TEST_RUN ...   Passed  167.73 sec

100% tests passed, 0 tests failed out of 1

Label Time Summary:
FRAMEWORK_pytest               = 167.73 sec*proc (1 test)
REQUIRES_gpu                   = 167.73 sec*proc (1 test)
SUITE_periodic                 = 167.73 sec*proc (1 test)
SUITE_periodic_REQUIRES_gpu    = 167.73 sec*proc (1 test)

Total Test time (real) = 168.03 sec
Failed to find key disabled in testsuite, continuing...

Test stability summary:
All tests were executed 1 times and passed 100% of the time.```
```

amzn-tommy (approved):

PR #165, opened 2021-07-08
Merged stabilization/2106 to development

cgalvan:
From upstream/stabilization/2106 commit 90e6f3b

No merge conflicts.

Signed-off-by: Chris Galvan chgalvan@amazon.com


Gem/Code/Source/Automation/ScriptManager.cpp
```diff
@@ -1164,9 +1164,9 @@ namespace AtomSampleViewer
 
     void ScriptManager::Script_CapturePassAttachment(AZ::ScriptDataContext& dc)
     {
-        if (dc.GetNumArguments() != 3)
+        if (dc.GetNumArguments() != 3 && dc.GetNumArguments() != 4)
```
carlitosan, line 1167: Nitpick, standard Lua is to allow extra arguments and just not do anything with them. So this should likely be written to only fuss if there are less than 3 arguments.

carlitosan (approved):

AMZN-daimini (approved):

PR #173, opened 2021-07-16
feat: Add build date to Git commit field for benchmark metadata.

synicalsyntax:
image


smurly (approved):
iso format date can be more useful for sort order YYYY-MM-DD, but it looks like this jut goes into a string with commit hash, so format isn't critical.

gadams3 (approved):

PR #182, opened 2021-07-28
Fix for w4267 ('var' : conversion from 'size_t' to 'type', possible loss of data)

amznestebanpapp:
https://docs.microsoft.com/en-us/cpp/error-messages/compiler-warnings/compiler-warning-level-3-c4267?view=msvc-160


rgba16f:
Please add what error W4267 is in the commit description

rgba16f (approved):

mrieggeramzn (approved):

PR #189, opened 2021-08-10
Fixing unicode calls required for https://github.com/o3de/o3de/pull/2757

amznestebanpapp:


Gem/Code/Source/Platform/Windows/Utils_Windows.cpp
```diff
@@ -22,19 +22,23 @@ namespace AtomSampleViewer
...
-
-            STARTUPINFO si;
+            AZStd::wstring exeW = L"C:\\Program Files\\Beyond Compare 4\\BCompare.exe";
+            AZStd::wstring filePathAW;
+            AZStd::to_wstring(filePathAW, filePathA.c_str());
+            AZStd::wstring filePathBW;
+            AZStd::to_wstring(filePathBW, filePathB.c_str());
+            AZStd::wstring commandLineW = L"\"" + exeW + L"\" \"" + filePathAW + L"\" \"" + filePathBW + L"\"";
+
+            STARTUPINFOW si;
```
lawsonamzn, line 32: wow paintful to see this burned into the EXE like this, but its not your fault.  Maybe open an issuee to redirect to regset file at least...

lawsonamzn (approved):

Gem/Code/Source/Platform/Windows/Utils_Windows.cpp
lumberyard-employee-dm: @o3de/sig-presentation What is this trying to do?
Is it to launch Beyond Compare for the user to see a UI?
Or is it actually trying to compare the file contents?
Because if it is the latter that can be done with two SystemFile objects and a `memcmp`

amznestebanpapp: same feeling... this is part of the ASV project so I figured it was ok for them...

rgba16f:
> Why this code launches Beyond Compare 4 is outside of the scope of this PR, so I see no issues with approving the changes.

We use it locally so we can see the visual difference between the golden image and our local run.

PR #195, opened 2021-08-17
fixing warnings that will be enabled by PR #3033

amznestebanpapp:


Gem/Code/Source/RHI/TextureMapExampleComponent.cpp
```diff
@@ -636,12 +636,12 @@ namespace AtomSampleViewer
...
-            indexBuffer[i + 2] = vertices + 1;
-            indexBuffer[i + 3] = vertices + 1;
-            indexBuffer[i + 4] = vertices + 3;
-            indexBuffer[i + 5] = vertices + 2;
+            indexBuffer[i + 0] = static_cast<uint16_t>(vertices + 0);
+            indexBuffer[i + 1] = static_cast<uint16_t>(vertices + 3);
+            indexBuffer[i + 2] = static_cast<uint16_t>(vertices + 1);
+            indexBuffer[i + 3] = static_cast<uint16_t>(vertices + 1);
+            indexBuffer[i + 4] = static_cast<uint16_t>(vertices + 3);
+            indexBuffer[i + 5] = static_cast<uint16_t>(vertices + 2);
```
rgba16f, line 644: why not just change vertices to be a uint16_t instead?

Gem/Code/Source/RHI/TriangleExampleComponent.cpp
```diff
@@ -214,9 +214,9 @@ namespace AtomSampleViewer
                 drawItem.m_arguments = drawIndexed;
                 drawItem.m_pipelineState = m_pipelineState.get();
                 drawItem.m_indexBufferView = &indexBufferView;
-                drawItem.m_shaderResourceGroupCount = RHI::ArraySize(shaderResourceGroups);
+                drawItem.m_shaderResourceGroupCount = static_cast<uint8_t>(RHI::ArraySize(shaderResourceGroups));
                 drawItem.m_shaderResourceGroups = shaderResourceGroups;
-                drawItem.m_streamBufferViewCount = static_cast<uint32_t>(m_streamBufferViews.size());
+                drawItem.m_streamBufferViewCount = static_cast<uint8_t>(m_streamBufferViews.size());
```
rgba16f, line 219: Should these changes where the type is narrowed from 32 bits to 8 be using aznumeric_cast instead?

Gem/Code/Source/RHI/TextureMapExampleComponent.cpp
amznestebanpapp: sure, I can change that

rgba16f (approved):

Gem/Code/Source/CullingAndLodExampleComponent.cpp
```diff
@@ -233,8 +233,8 @@ namespace AtomSampleViewer
             dirLightFP->SetViewFrustumCorrectionEnabled(handle, m_isCascadeCorrectionEnabled);
             dirLightFP->SetShadowFilterMethod(handle, s_shadowFilterMethods[m_shadowFilterMethodIndex]);
             dirLightFP->SetShadowBoundaryWidth(handle, m_boundaryWidth);
-            dirLightFP->SetPredictionSampleCount(handle, m_predictionSampleCount);
-            dirLightFP->SetFilteringSampleCount(handle, m_filteringSampleCount);
+            dirLightFP->SetPredictionSampleCount(handle, static_cast<uint16_t>(m_predictionSampleCount));
```
jeremyong-az, line 236: `m_predictionSampleCount` and `m_filteringSampleCount` definitely should be just `uint16_t` originally I think. The reason they aren't is because later code uses `ImGUI::SliderInt`. Instead of calling `ImGUI::SliderInt`, we should just call `ImGUI::SliderScalar` (and pass `ImGuiDataType_U16` as the second argument, all other arguments the same)

Gem/Code/Source/CullingAndLodExampleComponent.cpp
```diff
@@ -363,7 +363,7 @@ namespace AtomSampleViewer
                 ImGui::RadioButton("4", &m_cascadeCount, 4);
             if (cascadesChanged)
             {
-                m_directionalLightFeatureProcessor->SetCascadeCount(m_directionalLightHandle, m_cascadeCount);
+                m_directionalLightFeatureProcessor->SetCascadeCount(m_directionalLightHandle, static_cast<uint16_t>(m_cascadeCount));
```
jeremyong-az, line 366: `m_cascadeCount` can get the same treatment as `m_predictionSampleCount` and `m_filteringSampleCount`

Gem/Code/Source/RHI/TriangleExampleComponent.cpp
amznestebanpapp: I am going with `static_cast` instead because 
1) that is what is closer to what the compiler is doing; 
2) we found that aznumeric_cast usage across the board increases build time by ~5%
3) aznumeric_cast communicates that the casting is intentional, but I want that decision to be made by each team. Teams could choose to do something different. E.g. EMotionFX refactored their types to reduce casting

Gem/Code/Source/RHI/TextureMapExampleComponent.cpp
```diff
@@ -636,12 +636,12 @@ namespace AtomSampleViewer
         int vertices = 0;
         for (int i = 0; i < arraySize; i += 6)
         {
-            indexBuffer[i + 0] = vertices + 0;
-            indexBuffer[i + 1] = vertices + 3;
-            indexBuffer[i + 2] = vertices + 1;
-            indexBuffer[i + 3] = vertices + 1;
-            indexBuffer[i + 4] = vertices + 3;
-            indexBuffer[i + 5] = vertices + 2;
+            indexBuffer[i + 0] = static_cast<uint16_t>(vertices + 0);
```
jeremyong-az, line 639: `vertices` can be declared as a `uint16_t` here (and rename it to `indexCursor` or something please while you're here 😉)

Gem/Code/Source/ShadowExampleComponent.cpp
```diff
@@ -590,12 +590,12 @@ namespace AtomSampleViewer
                 if (ScriptableImGui::SliderInt("Prediction # ##Directional", &m_predictionSampleCountDirectional, 4, 16))
                 {
                     m_directionalLightFeatureProcessor->SetPredictionSampleCount(
-                        m_directionalLightHandle, m_predictionSampleCountDirectional);
+                        m_directionalLightHandle, static_cast<uint16_t>(m_predictionSampleCountDirectional));
```
jeremyong-az, line 593: Similar changes here with respect to the various sample counts if you would

jeremyong-az (changes requested):

Gem/Code/Source/AreaLightExampleComponent.cpp
```diff
@@ -437,7 +437,7 @@ namespace AtomSampleViewer
         for (uint32_t i = 0; i < vertexCount; ++i)
         {
             // Get a point on the circle and scale it by the min or max radius for every other point.
-            AZ::Vector3 point = GetCirclePoint(i, vertexCount) * minMaxRadius[i % 2];
+            AZ::Vector3 point = GetCirclePoint(static_cast<float>(i), static_cast<float>(vertexCount)) * minMaxRadius[i % 2];
```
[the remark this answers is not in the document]
amznestebanpapp: sorry, I cannot do those sort of changes for every case... if I were to go down that rabbit hole I would never be able to get this done. That should be up to the team that owns the code
Same for the comments below. 
The approach I took for these changes is to do the least invasive change possible (in this case I am doing what the compiler is doing) and let the teams clean it up.

jeremyong-az (approved):
Per conversation with @amznestebanpapp the more invasive cleanup tasks will get tackled later either by myself or someone on the Atom team. Thanks Esteban!

PR #197, opened 2021-08-17
MSAA sample is missing lighting when viewing the non-MSAA mode

dmcdiarmid-ly:
Added SampleComponentManagerBus events for setting the number of MSAA samples.
Broadcasted these events in the MSAA_RPI_ExampleComponent to properly handle changing the pipeline to a different number of samples.

Signed-off-by: dmcdiar <dmcdiar@amazon.com>

### DIFF
--- a/Gem/Code/Source/SampleComponentManager.cpp
+++ b/Gem/Code/Source/SampleComponentManager.cpp
@@ -136,6 +136,11 @@ namespace AtomSampleViewer
         const char* TransientAttachmentProfilerToolName = "Transient Attachment Profiler";
     }
 
+    bool IsValidNumMSAASamples(int numSamples)
+    {
+        return (numSamples == 1) || (numSamples == 2) || (numSamples == 4) || (numSamples == 8);
+    }
+
     SampleEntry SampleEntry::NewRHISample(const AZStd::string& name, const AZ::Uuid& uuid)
     {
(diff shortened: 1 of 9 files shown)


moudgils (approved):

Gem/Code/Source/SampleComponentManager.cpp
```diff
@@ -1490,7 +1503,10 @@ namespace AtomSampleViewer
         pipelineDesc.m_name = "RPISamplePipeline";
         pipelineDesc.m_rootPassTemplate = GetRootPassTemplateName();
         pipelineDesc.m_mainViewTagName = "MainCamera";
-        pipelineDesc.m_renderSettings.m_multisampleState.m_samples = GutNumMSAASamples();
+
+        // set pipeline MSAA samples
+        AZ_Assert(m_numMSAASamples >= 1 && m_numMSAASamples <= 8, "Invalid MSAA sample setting");
```
invertednormal, line 1508: Do we need to check for specific numbers here... like only 1,2,4 and 8 are valid right?

invertednormal (approved):

Gem/Code/Source/SampleComponentManager.cpp
dmcdiarmid-ly: Good idea, added.

PR #202, opened 2021-08-20
Remove periodic test suite registrations.

jromnoa:
- We need GPU nodes for all of these tests and only main suite uses them. This deletes redundant suites we won't need for ASV AR in the future.


jromnoa:
Was able to build and run the tests successfully.

atomsampleviewer_asset_files.cmake
```diff
@@ -46,8 +46,7 @@ set(FILES
     Scripts/ShadowTest.bv.lua
     Scripts/StreamingImageTest.bv.lua
     Scripts/TransparentTest.bv.lua
-    Scripts/_AutomatedPeriodicTestSuite_.bv.lua
-    Scripts/_AutomatedReviewTestSuite_.bv.lua
+	Scripts/_AutomatedReviewWARPTestSuite_.bv.lua
```
smurly, line 49: this seems to have a tab character for indent. not sure if that's an issue

smurly (approved):

evanchia-ly-sdets (approved):

atomsampleviewer_asset_files.cmake
jromnoa: ill convert to spaces

PR #211, opened 2021-08-30
Fix compile errors after warning updates

spham-amzn:


mrieggeramzn (approved):

spham-amzn:
Looks like this is a dup of https://github.com/o3de/o3de-atom-sampleviewer/pull/208

PR #213, opened 2021-09-01
Fixes compilation bug:

galibzon:
D:/GIT/o3de/AtomSampleViewer/Gem/Code/Source/Automation/ScriptReporter.cpp(1106,74):
error C2440: 'initializing': cannot convert from
'AZ::IO::FixedMaxPathString' to
'AZStd::basic_string<char,AZStd::char_traits<Element>,AZStd::allocator>'

const AZStd::string projectPath = AZ::Utils::GetProjectPath();

Signed-off-by: garrieta <garrieta@amazon.com>

### DIFF
--- a/Gem/Code/Source/Automation/ScriptReporter.cpp
+++ b/Gem/Code/Source/Automation/ScriptReporter.cpp
@@ -1103,7 +1103,7 @@ namespace AtomSampleViewer
             const AZStd::string screenshotWarningsLogLine = AZStd::string::format("Screenshot warnings: %u \n", scriptReport.m_screenshotWarningCount);
             const AZStd::string failedScreenshotsLogLine = "\nScreenshot test info below.\n";     
 
-            const AZStd::string projectPath = AZ::Utils::GetProjectPath();
+            const auto projectPath = AZ::Utils::GetProjectPath();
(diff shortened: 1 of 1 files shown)


Gem/Code/Source/Automation/ScriptReporter.cpp
```diff
@@ -1103,7 +1103,7 @@ namespace AtomSampleViewer
             const AZStd::string screenshotWarningsLogLine = AZStd::string::format("Screenshot warnings: %u \n", scriptReport.m_screenshotWarningCount);
             const AZStd::string failedScreenshotsLogLine = "\nScreenshot test info below.\n";     
 
-            const AZStd::string projectPath = AZ::Utils::GetProjectPath();
```
rgba16f, line 1106: should this be a reference or is it intentionally copying the string?

rgba16f (approved):

Gem/Code/Source/Automation/ScriptReporter.cpp
galibzon: In this case it doesn't have to be reference. A copy is just fine.

galibzon: In other words, copy elision in action.

jromnoa (approved):

PR #218, opened 2021-09-13
[RHI][Vulkan] - Have different padding size due to platform differences.

AMZN-tpeng:
Signed-off-by: Peng <tonypeng@amazon.com>


Shaders/RHI/TrianglesConstantBuffer.azsl
```diff
@@ -21,14 +21,32 @@
 
 #endif
 
+
+#if AZ_TRAIT_CONSTANT_BUFFER_ALIGNMENT <= 64
```
moudgils, line 25: Where is AZ_TRAIT_CONSTANT_BUFFER_ALIGNMENT defined?

mrieggeramzn, line 25: Should this be 16?
I'm thinking that with no padding, the size of that struct is 80 bytes. 64 and 32 don't divide equally into 80 bytes, but 16 bytes does.

AMZN-tpeng: They are in Gems/Atom/Asset/Shader/Code/AZSL/Platform/Android/Vulkan/AzslcHeader.azsli in ASV

moudgils: if AZ_TRAIT_CONSTANT_BUFFER_ALIGNMENT == 64 it should be 3 and if AZ_TRAIT_CONSTANT_BUFFER_ALIGNMENT == 16 it should be 0

jiaweig-amzn: I don't think an if-else statement is needed. They have solid mathematical relations. We can do:
#define AZ_TRAIT_CONSTANT_BUFFER_ALIGNMENT_IN_FLOAT4
(AZ_TRAIT_CONSTANT_BUFFER_ALIGNMENT / 16)
#define NUM_INSTANCE_INFO_PADDINGS
(AZ_TRAIT_CONSTANT_BUFFER_ALIGNMENT_IN_FLOAT4 + (5 / AZ_TRAIT_CONSTANT_BUFFER_ALIGNMENT_IN_FLOAT4 
 * AZ_TRAIT_CONSTANT_BUFFER_ALIGNMENT_IN_FLOAT4)  - 5)

In short, if the current data size is A (size in float4, here is 5), the alignment is B (size in float4), then the padding is
B + (A / B) * B - A

Shaders/RHI/TrianglesConstantBuffer.azsl
```diff
@@ -21,14 +21,32 @@
...
+
+#define NUM_INSTANCE_INFO_PADDINGS 0
```
jiaweig-amzn, line 27: If result in 0, does float4 m_padding0[0]; compile?

Shaders/RHI/TrianglesConstantBuffer.azsl
AMZN-tpeng: m_padding0[0] would not compile, thus the excessive #if statements. 

This would have been more concise with if so: 

m_padding0[(AZ_TRAIT_CONSTANT_ALIGNMENT-(4*16+16)) % AZ_TRAIT_CONSTANT_ALIGNMENT)];

AMZN-tpeng: Added more alignment sizes to handle the different cases.

mrieggeramzn (approved):

moudgils (approved):

PR #223, opened 2021-09-23
Adding ASV decal normal maps

mrieggeramzn:
Adding several default normal maps to the decal samples in AtomSampleViewer
Adding light to the decal ASV sample
Updating decal xshot
ATOM-4689

image

Signed-off-by: mrieggeramzn <mriegger@amazon.com>


antonmic (approved):

Gem/Code/Source/DecalExampleComponent.cpp
```diff
@@ -158,4 +182,14 @@ namespace AtomSampleViewer
         m_decalContainerClone = AZStd::make_unique<DecalContainer>(decalFeatureProcessor, AZ::Vector3(-1,0,0));
     }
 
+    void DecalExampleComponent::UpdateDirectionalLight()
+    {
+        using namespace AZ;
+
+        constexpr float directionalLightDist = 10.0f;
```
Adi-Amazon, line 189: nit: use this instead of the 10.0?

Adi-Amazon (approved):

Gem/Code/Source/DecalExampleComponent.cpp
mrieggeramzn: @Adi-Amazon Hi Adi. I missed your comment until just now. Could you clarify what you meant?

PR #234, opened 2021-10-20
Show path to exported test results in ASV.

jromnoa:
- Adds a display for where the exported test results are saved to in ASV when the "Export Test Results" button is clicked.


Gem/Code/Source/Automation/ScriptReporter.h
```diff
@@ -88,6 +88,10 @@ namespace AtomSampleViewer
         // For exporting test results
         void ExportTestResults();
 
+        // Path to the exported test results file.
+        AZStd::string m_exportedTestResultsPath = "Click the 'Export Test Results' button.";
+        AZStd::string GetExportedTestResultsPath() const;
```
hershey5045, line 93: When adding member functions and variables, it's important to determine its access of level. Best practice is to expose functions/variables on an as-needed basis. So essentially, only the ScriptReporter class needs to access these, so declaring them as public is too permissive. Please move them down below the private section. Lastly, be sure to group the member function and member variable separately (follow the formatting within this class)

hershey5045, line 93: Can we rename this to GenerateExportedTestResultsPath() since we are essentially creating a unique filename dynamically?

hershey5045 (approved):
Looks good!

PR #260, opened 2021-11-02
Add Soak Test To Validate reload notifications of Shader related assets

galibzon:
Added new Example called "RPI/ShaderReloadSoakTest".
As an example it provides an ImGUI UI to change the
shader used for the FullscreenTrianglePass which renders
a solid Red, Green or Blue color.

A button to check the expected color is available.

A lua script is available: ShaderReloadSoakTest.bv.lua, which
mimics pressing the buttons and testing for the expected color
when the shader asset changes.

Because this is a new Soak Test it should NOT be added
to any of the automated test suites.

Signed-off-by: galibzon <66021303+galibzon@users.noreply.github.com>


galibzon:
ASV AR Green: https://jenkins-pipeline.agscollab.com/view/all/job/o3de-atom-sampleviewer-dev/job/Atom%252Fgalibzon%252FATOM-5134%252Fshader_reload_soak_test/1/

Gem/Code/Source/SampleComponentManager.cpp
```diff
@@ -282,6 +283,7 @@ namespace AtomSampleViewer
         SampleComponentManager::RegisterSampleComponent(SampleEntry::NewRPISample( "RPI/SceneReloadSoakTest", azrtti_typeid<SceneReloadSoakTestComponent>() ));
         SampleComponentManager::RegisterSampleComponent(SampleEntry::NewRPISample( "RPI/Shading", azrtti_typeid<ShadingExampleComponent>() ));
         SampleComponentManager::RegisterSampleComponent(SampleEntry::NewRPISample( "RPI/StreamingImage", azrtti_typeid<StreamingImageExampleComponent>() ));
+        SampleComponentManager::RegisterSampleComponent(SampleEntry::NewRPISample( "RPI/ShaderReloadSoakTest", azrtti_typeid<ShaderReloadSoakTestComponent>() ));
```
santorac, line 286: I don't think the sample should be called "ShaderReloadSoakTest" because the sample itself doesn't soak anything. It just provides buttons to apply shader changes. I would just call it "ShaderReloadTest" or "FullscreenShaderReloadTest". You can still call the _script_ "ShaderReloadSoakTest.bv.lua" because that does do soaking.

Gem/Code/Source/ShaderReloadSoakTestComponent.cpp
```diff
@@ -0,0 +1,322 @@
...
+        }
+
+        AzFramework::StringFunc::Path::Join(mainTestFolder.c_str(), "Temp/", m_tempSourceFolder);
+        if (!io->CreatePath(m_tempSourceFolder.c_str()))
+        {
+            AZ_Error("ShaderReloadSoakTestComponent", false, "Could not create temp folder '%s'.", m_tempSourceFolder.c_str());
+        }
+    }
+
+    void ShaderReloadSoakTestComponent::CopyTestFile(const char * originalName, const char * newName, bool replaceIfExists)
```
santorac, line 60: Can any of this class be factored out with MaterialHotReloadTestComponent? Maybe form a common base class?

Gem/Code/Source/ShaderReloadSoakTestComponent.cpp
```diff
@@ -0,0 +1,322 @@
...
+        AZ::TickBus::Handler::BusConnect();
+
+        InitTestDataFolders();
+        CopyTestFile(RedShaderFile, "Fullscreen.azsl");
+        CopyTestFile("Fullscreen.shader.txt", "Fullscreen.shader");
+        m_expectedPixelColor = RED_COLOR;
+
+        // A short wait is necessary as the pipeline creation will fatally fail
+        // if the "Fullscreen.azshader" doesn't exist.
+        AZStd::this_thread::sleep_for(AZStd::chrono::milliseconds(1000));
```
santorac, line 141: Instead of sleeping, can you look for signals from the AP about the asset status? See the AssetStatusTracker class for some ideas.

Scripts/ShaderReloadSoakTest.bv.lua
```diff
@@ -0,0 +1,38 @@
...
+    SetImguiValue(buttonName, true)
+    IdleSeconds(shaderReloadWaitSeconds) -- Idle for a bit to give time for the shader to reload
+    SetShowImGui(false)
+    SetImguiValue('Check color', true)
+    IdleSeconds(pixelCaptureWaitSeconds)
+    SetShowImGui(true)
+end
+
+local shaderReloadWaitSeconds = 3
+local pixelCaptureWaitSeconds = 1
```
santorac, line 29: Did you try using the AssetTracking_* functions like MaterialHotReloadTest? Having arbitrary timeouts makes things take longer than necessary, and it doesn't scale well across platforms where shader compilation can take much longer (like on Mac IIRC).

santorac (changes requested):

Gem/Code/Source/SampleComponentManager.cpp
galibzon: Good point.

Gem/Code/Source/ShaderReloadSoakTestComponent.cpp
galibzon: I was thinking about all that, but I rather create a ticket for post reinvent.

Gem/Code/Source/ShaderReloadSoakTestComponent.cpp
galibzon: I was just being lazy. We'll try to use the AssetCollectionAsyncLoader or AssetStatusTracker

Scripts/ShaderReloadSoakTest.bv.lua
galibzon: Will take a look.

Gem/Code/Source/ShaderReloadSoakTestComponent.cpp
Adi-Amazon: +1

Scripts/ShaderReloadSoakTest.bv.lua
```diff
@@ -0,0 +1,38 @@
...
+
+
+-- Fixme. As a Soak Test, this should run for a long time and exit
+-- on the first failure, or upon user request.
+for i=1,5 do
+    TestButton("Red shader", shaderReloadWaitSeconds, pixelCaptureWaitSeconds)
```
Adi-Amazon, line 35: It's a very good test but how realistic is the assumption that it will always finish reload in 3 seconds?
Also, running 15 cycles of it seems like a long test (>15 seconds), is there away to bring it to 'fail conditions' and test it at that?

Adi-Amazon (approved):
Not blocking the submission, but two main observations:
1. timer to be move to depend on signal
2. The test is rather long and doesn't guarantee that the shader finished reloading - can we somehow shorten it and better guarantee load finish + fail conditions?

Scripts/ShaderReloadSoakTest.bv.lua
galibzon: I'm going to add some better notification trackings to reduce the wait time. The Lua API is insufficient to do what you mention, which I already thought about it too: 1- Ability to exit on first failure. 2- Ability to exit the test at any time.

galibzon:
@Adi-Amazon & @santorac This new version uses AssetStatusTracker to avoid fuzzy wait times around shader recompilation.

Gem/Code/Source/ShaderReloadTestComponent.cpp
```diff
@@ -122,22 +126,51 @@ namespace AtomSampleViewer
...
+        {
+            AZStd::this_thread::sleep_for(AZStd::chrono::milliseconds(MilliWaits));
+            timeWaitMillis += MilliWaits;
+            if (timeWaitMillis >= MaxWaitTimeMillis)
+            {
+                AZ_Error(LogName, false, "Failed to activate this test because exceeded wait time of %u milliseconds", MaxWaitTimeMillis);
+                return;
+            }
+        }
+        assetStatusTracker.StopTracking();
```
santorac, line 169: It's a bummer to block the whole UI while waiting for the shader to compile. Can this be done in an async way? You could make the AssetStatusTracker be a member and check DidExpectedAssetsFinish() in OnTick instead.

Gem/Code/Source/ShaderReloadTestComponent.cpp
```diff
@@ -122,22 +126,51 @@ namespace AtomSampleViewer
...
+
+        // The above wait time is just for shader recompilation, it is important to wait a little bit more
+        // for the shader asset to be discoverable when the FullscreenTriangle pass is loaded and tries
+        // to reference the shader. 1 second is plenty and generous.
         AZStd::this_thread::sleep_for(AZStd::chrono::milliseconds(1000));
```
santorac: Can you check for the presence of the shader asset in the asset manager instead of just sleeping? In MaterialHotReloadTest we do this by checking AssetCatalogRequestBus GetAssetIdByPath (see MaterialHotReloadTestComponent::GetAssetId). And this could be done in OnTick to keep the UI responsive.

Gem/Code/Source/ShaderReloadTestComponent.cpp
```diff
@@ -122,22 +126,51 @@ namespace AtomSampleViewer
...
 
-    void ShaderReloadSoakTestComponent::Activate()
+    void ShaderReloadTestComponent::Activate()
     {
         AZ::TickBus::Handler::BusConnect();
 
         InitTestDataFolders();
+
+        DeleteTestFile("Fullscreen.shader");
+        DeleteTestFile("Fullscreen.azsl");
```
santorac, line 141: In MaterialHotReloadTest I had to wait for the assets to fully disappear from the asset manager before continuing. Is that not needed here? See MaterialHotReloadTestComponent's InitStatus::ClearingTestAssets.

Gem/Code/Source/ShaderReloadTestComponent.cpp
santorac: Or factor out MaterialHotReloadTestComponent::GetTestAssetStatus

galibzon:
New AR Run is Green: https://jenkins-pipeline.agscollab.com/view/all/job/o3de-atom-sampleviewer-dev/job/Atom%252Fgalibzon%252FATOM-5134%252Fshader_reload_soak_test/3/

Scripts/ShaderReloadSoakTest.bv.lua
```diff
@@ -12,27 +12,36 @@
...
+    -- Even though the shader has been recompiled, it takes a few frames
+    -- for the notification to bubble up and get the screen refreshed.
+    IdleSeconds(0.25)
+
     SetShowImGui(false)
     SetImguiValue('Check color', true)
-    IdleSeconds(pixelCaptureWaitSeconds)
+
+    -- Need a few frames to capture the screen and compare expected color.
+    IdleSeconds(0.25)
```
santorac, line 32: This seems small, but when you do it many times it adds up (3.75s in this case). You could probably leverage ScriptRunnerRequests::PauseScript/ResumeScript in the component to pause while the screenshot is being taken.

Actually, why not use the screenshot comparison process that we use in other tests? Then you could remove the Check Color button, and all the bespoke frame capture code you have in the component.

santorac (changes requested):
I see renames like "...loadSoakTest/TestData/Fullscreen_RED.azsl → ...ers/ShaderReloadTest/Temp/Fullscreen.azsl". It looks like an accidental "git add".

Gem/Code/Source/ShaderReloadSoakTestComponent.cpp
santorac: I expect my comments below about UI responsiveness and timing would be nicely alleviated by factoring out with MaterialHotReloadTestComponent

Gem/Code/Source/ShaderReloadTestComponent.cpp
galibzon: The clean solution is to use AsyncAssetLoader.

Gem/Code/Source/ShaderReloadTestComponent.cpp
galibzon: will use AsyncAssetLoader

Gem/Code/Source/ShaderReloadTestComponent.cpp
galibzon: The way AssetStatusTracker works, it may be necessary to guarantee that the assets are fully deleted, but I can do that check during Activate(). The idea is that you can stop the debugger and never call Deactivate(), so next time around if this case is not considered inside Activate() it will be a guaranteed failure.

Scripts/ShaderReloadSoakTest.bv.lua
galibzon: Will do that after reinvent. But I think We need other APIs for tests that don;t require full screen capture comparison.

santorac (approved):
Marking as approved with the understanding that comments about timing and refactoring will be addressed in a followup PR after we get through the upcoming release.

PR #269, opened 2021-11-05
ATOM-16803 AtomSampleViewerStandalone "CheckerboardTest.bv.lua" test …

VickyAtAZ:
…script hard locks then crashes.

Signed-off-by: Qing Tao <55564570+VickyAtAZ@users.noreply.github.com>


santorac (approved):

galibzon (approved):
When doing ShaderReloadTestComponent,  I did the same but it was causing a pesky warning everytime the example was opened. So I left an empty implementation. Can you confirm there are no polluting warnings?

invertednormal (approved):

PR #270, opened 2021-11-05
Atom/qingtao/get default scene stab asv

VickyAtAZ:
Cherry picked from commit 69d726d97699feb63de78c1391ca88928ea205c1 [69d726d] in development branch

* ATOM-16747 Replace all the usage of RPISystemInterface::GetDefaultScene in ASV.

Signed-off-by: Qing Tao <55564570+VickyAtAZ@users.noreply.github.com>

### DIFF
--- a/Gem/Code/Source/AreaLightExampleComponent.cpp
+++ b/Gem/Code/Source/AreaLightExampleComponent.cpp
@@ -63,16 +63,15 @@ namespace AtomSampleViewer
     void AreaLightExampleComponent::Activate()
     {
         // Get Feature processors
-        AZ::RPI::ScenePtr scene = AZ::RPI::RPISystemInterface::Get()->GetDefaultScene();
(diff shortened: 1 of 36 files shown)


gadams3 (approved):

galibzon (approved):

Adi-Amazon (approved):
Approved since it passed already and looks like a good change.
My only question is: in the previous implementation 'scene' was acquired in many occasions every frame, hence securing validity and existence of the scene.  Are you sure that acquiring 'm_scene' once, it'll always be valid with the new implementation?

VickyAtAZ:
> Approved since it passed already and looks like a good change.
> My only question is: in the previous implementation 'scene' was acquired in many occasions every frame, hence securing validity and existence of the scene. Are you sure that acquiring 'm_scene' once, it'll always be valid with the new implementation?

Yes, the scene is set when ever a sample entity is created and stay valid when this sample is active. It may change when switching sample which the implementation can handle that.

PR #272, opened 2021-11-08
Updating AtomSampleViewerApplication to the TimeSystem change

amzn-sean:
This change is required with this PR in o3de.
https://github.com/o3de/o3de/pull/5409

### DIFF
--- a/Gem/Code/Source/Automation/ScriptManager.cpp
+++ b/Gem/Code/Source/Automation/ScriptManager.cpp
@@ -394,7 +394,7 @@ namespace AtomSampleViewer
 
                 if (m_frameTimeIsLocked)
                 {
-                    AZ::Interface<AZ::IConsole>::Get()->PerformCommand("t_frameTimeOverride 0");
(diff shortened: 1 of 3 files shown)


moraaar (approved):
LGTM

santorac (changes requested):
The changes look fine. Please be sure to run AtomSampleViewer's automated tests. 

There is some special sauce in ASV that sort of hacks the tick system to ensure deterministic timing for certain automated tests. So those scripts need to pass to make sure things still work as expected. So at a minimum, these scripts need to be tested, and I would run them a few times just to make sure....

- DynamicMaterialTest.bv.lua
- SceneReloadSoakTest.bv.lua

Also, it's good practice to run the whole ASV test suite before merging...

- \_FullTestSuite_.bv.lua

And we are now being asked to run ASV an jenkins before committing ASV changes. I'll send you a link about that, since it isn't published publicly yet.

santorac (approved):

PR #286, opened 2021-11-18
Add a HighInstance test that by default uses a 10k object entity grid as a basic perf test

rgba16f:

### DIFF
--- a/Gem/Code/Source/HighInstanceExampleComponent.cpp
+++ b/Gem/Code/Source/HighInstanceExampleComponent.cpp
@@ -18,6 +18,7 @@
 
 #include <AzCore/Debug/EventTrace.h>
 #include <AzCore/Serialization/SerializeContext.h>
+#include <AzFramework/Windowing/WindowBus.h>
 
 #include <RHI/BasicRHIComponent.h>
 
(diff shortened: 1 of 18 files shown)


Gem/Code/Source/HighInstanceExampleComponent.cpp
```diff
@@ -0,0 +1,309 @@
...
+                ->Version(0)
+                ;
+        }
+    }
+
+
+    HighInstanceTestComponent::HighInstanceTestComponent() 
+        : m_materialBrowser("@user@/AssetLoadTestComponent/material_browser.xml")
+        , m_modelBrowser("@user@/AssetLoadTestComponent/model_browser.xml")
+        , m_imguiSidebar("@user@/AssetLoadTestComponent/sidebar.xml")
```
VickyAtAZ, line 45: The paths need update.

Gem/Code/Source/HighInstanceExampleComponent.h
```diff
@@ -0,0 +1,93 @@
...
+#include <Utils/ImGuiAssetBrowser.h>
+#include <AzCore/Component/TickBus.h>
+
+namespace AtomSampleViewer
+{
+    /*
+        This test loads a configurable lattice of entities with randomized meshes and materials.
+
+        The assets that are applied to the entities are chosen from a configurable 
+        "allow-list" of models and materials.
```
VickyAtAZ, line 22: This comment seems need to be updated to reflect the purpose of this sample. Otherwise it doesn't have much different than the asset load test sample.

VickyAtAZ (commented):
are you planning to add lua test for this new example?

galibzon (changes requested):
Can you just increase the limits of the knobs (sliders) in AssetLoadTestComponent and not having to add another sample to ASV?

Gem/Code/Source/HighInstanceExampleComponent.cpp
```diff
@@ -0,0 +1,309 @@
...
+            return AZ::RPI::AssetUtils::GetAssetIdForProductPath(DefaultPbrMaterialPath, AZ::RPI::AssetUtils::TraceLevel::Error);
+        }
+    }
+
+
+    void HighInstanceTestComponent::OnTick([[maybe_unused]] float deltaTime, [[maybe_unused]] AZ::ScriptTimePoint scriptTime)
+    {
+        AZ_TRACE_METHOD();
+
+        if (m_updateTransformEnabled)
```
moudgils, line 282: I wonder (given the reason for this sample is profilling) if it makes sense to introduce a sleep here for the same amount of time as we think a modern open AAA game takes for gameplay. Basically we dont have a sample that stress tests gameplay/sim and it maybe a good idea to simulate that time here. It will then force us to think about moving Rendering to another thread sooner rather than later.

rgba16f: Increasing the range of the sliders works for most of the Atom team, but the request here was to have something that any QA representative could trivially run and see how Atom's performance has changed. So the aim is to have a simple test that stresses Atoms cpu (not rendering) performance independent of anything else.

rgba16f:
> Can you just increase the limits of the knobs (sliders) in AssetLoadTestComponent and not having to add another sample to ASV?

Increasing the range of the sliders works for most of the Atom team, but the request here was to have something that any QA representative could trivially run and see how Atom's performance has changed. So the aim is to have a simple test that stresses Atoms cpu (not rendering) performance independent of anything else.

Gem/Code/Source/HighInstanceExampleComponent.cpp
rgba16f: This test is wholly about Atom's cpu perf. I will make a separate test in the AutomatedTesting project that pushes how Atom performs as part of O3DE

galibzon (approved):

VickyAtAZ (approved):

PR #287, opened 2021-11-18
AR Fixes for project specific artifact collection - Cherrypick from `Stabilization/2110` (#285)

amzn-changml:
Signed-off-by: Mike Chang <changml@amazon.com>

### DIFF
--- a/Scripts/build/Jenkins/Jenkinsfile
+++ b/Scripts/build/Jenkins/Jenkinsfile
@@ -460,7 +460,7 @@ def ExportTestResults(Map options, String platform, String type, String workspac
     catchError(message: "Error exporting tests results (this won't fail the build)", buildResult: 'SUCCESS', stageResult: 'FAILURE') {
         def o3deroot = "${workspace}/${ENGINE_REPOSITORY_NAME}"
         dir("${o3deroot}/${params.OUTPUT_DIRECTORY}") {
-            junit testResults: "Testing/**/*.xml"
+            junit testResults: "Testing/**/*.xml", skipPublishingChecks: true
             palRmDir("Testing")
             // Recreate test runner xml directories that need to be pre generated
             palMkdir("Testing/Pytest")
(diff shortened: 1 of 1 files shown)


Scripts/build/Jenkins/Jenkinsfile
```diff
@@ -498,12 +497,12 @@ def UploadAPLogs(Map options, String branchName, String platformName, String job
             }
             def command = "${pythonPath} -u ${s3UploadScriptPath} --base_dir ${apLogsPath} " +
                           "--file_regex \".*\" --bucket ${env.AP_LOGS_S3_BUCKET} " +
-                          "--search_subdirectories True --key_prefix ${env.JENKINS_JOB_NAME}/${branchName}/${env.BUILD_NUMBER}/${platformName}/${jobName} " +
+                          "--search_subdirectories True --key_prefix ${PROJECT_REPOSITORY_NAME}/${branchName}/${env.BUILD_NUMBER}/${platformName}/${jobName} " +
```
shirangj, line 500: I don't see where PROJECT_REPOSITORY_NAME is set

amzn-changml: It's a const at the top of the file: https://github.com/o3de/o3de-atom-sampleviewer/blob/c865eaf5eead9061f2b6e188383c5dd824110732/Scripts/build/Jenkins/Jenkinsfile#L16

amzn-changml: I put it in place to ensure that the project job doesn't pollute the same space with the o3de jobs, since the main branches are all named the same

shirangj: Ok, env.JENKINS_JOB_NAME is the name of the Jenkins job, since ASV job name is different than o3de jobs, it won't pollute the o3de folder, you can see ASV artifacts are in o3de-atom-sampleviewer folder.
The problem of using PROJECT_REPOSITORY_NAME is that if we have two ASV Jenkins jobs, these two jobs will upload artifacts to the same place.

amzn-changml: Actually, the current artifacts for o3de-atom-sampleviewer folder are there because I tested this with this code.

shirangj: Hmmm... Did you run with env.JENKINS_JOB_NAME too? I cannot any ASV AP logs except "o3de-atom-sampleviewer" on S3...

amzn-changml: I did, the result was `null`. I'll give it a run with the old parameter so we can validate this

lumberyard-employee-dm (approved):

cgalvan (approved):

amzn-changml:
Closing this as it was merged into development as of b437a28

PR #290, opened 2021-11-19
ATOM-16177 [RHI][Android] Light Culling sample crashes for device lost

jiaweig-amzn:
Reduce the initial point light number to 10

### DIFF
--- a/Gem/Code/Source/LightCullingExampleComponent.cpp
+++ b/Gem/Code/Source/LightCullingExampleComponent.cpp
@@ -38,13 +38,15 @@
 #include <Atom/RPI.Reflect/Asset/AssetUtils.h>
 #include <RHI/BasicRHIComponent.h>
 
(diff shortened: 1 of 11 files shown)


AMZN-tpeng (approved):

moudgils (commented):
So we actually need Sponza for Android. I dont think we can submit this change. We are using Sponza for perf experiments on other devices so we can't just replace it. Personally I would like a way to confirm the hypothesis you have mentioned. 

For example, We can ask Jonny to create LODs for Sponza scene and see if using lower LODs help fix this issue. Its surprising to me that Pixel 5 which is near a high end device is unable to render the default Sponza scene. Another thing we can try is disabling shadow or simplifying forward pass drastically to see if it helps resolve the issue to isolate which pass is causing the crash. We can also think about reaching out to vendors with this sample to see if they can help us isolate the issue.

PR #299, opened 2021-12-09
Redcode/driller removal

amznestebanpapp:
- Adds changes from the last project template (improves compile detection in Linux)
-  Removes FileIOErrorHandler. This was using `FileIOEventBus` which was requested to be removed by @lawsonamzn since it can produce error codes from other file operations to be reported (the bus does not handle threading and/or is not IDed by operation). If error code operations are needed, we will need to pass them through the stack (a parameter in the methods being called to perform the file operations) and/or create a thread-aware error code retrieval. 
Removing to unblock.
- EventTrace was removed since it was relying on the driller which was not doing anything. Replaced its usage with the new profiler macros.
- Removed the inclusion of gridmate's driller files that were not used and were removed.

Also noted that ASV is setting up GridMate but not doing anything with it. That seems to be something that can be cleaned up (not doing as part of this change). Note GridMate is also going to be deprecated/removed at some point.


rgba16f (approved):

Gem/Code/Source/HighInstanceExampleComponent.cpp
```diff
@@ -309,7 +309,7 @@ namespace AtomSampleViewer
 
     void HighInstanceTestComponent::OnTick([[maybe_unused]] float deltaTime, [[maybe_unused]] AZ::ScriptTimePoint scriptTime)
     {
-        AZ_TRACE_METHOD();
+		AZ_PROFILE_FUNCTION(AtomSampleViewer);
```
AMZN-ScottR, line 312: nit: odd spacing here

AMZN-ScottR (approved):

galibzon (approved):

PR #302, opened 2021-12-10
Streamline ASV sample registration

jeremyong-az:
ASV samples are now registered in a single location and the ASV module
will pull descriptors from this single location. In addition, ASV
samples are now grouped in nested submenus in ImGUI.

UI change preview:

image

### DIFF
--- a/Gem/Code/Source/SampleComponentManager.cpp
+++ b/Gem/Code/Source/SampleComponentManager.cpp
@@ -228,9 +228,9 @@ namespace AtomSampleViewer
         dependent.push_back(AZ_CRC("AzFrameworkConfigurationSystemComponentService", 0xcc49c96e)); // Ensures a scene is created for the GameEntityContext
     }
 
-    const AZStd::vector<SampleEntry>& SampleComponentManager::GetSamples()
+    const AZStd::array_view<SampleEntry> SampleComponentManager::GetSamples()
     {
-        static AZStd::vector<SampleEntry> sampleEntries{
+        static SampleEntry sampleEntries[] = {
             NewRHISample<AlphaToCoverageExampleComponent>("AlphaToCoverage"),
             NewRHISample<AsyncComputeExampleComponent>("AsyncCompute"),
             NewRHISample<BindlessPrototypeExampleComponent>("BindlessPrototype", []() {return Utils::GetRHIDevice()->GetFeatures().m_unboundedArrays; }),
@@ -293,7 +293,7 @@ namespace AtomSampleViewer
             NewFeaturesSample<TransparencyExampleComponent>("Transparency"),
         };
 
-        return sampleEntries;
+        return {sampleEntries, AZ_ARRAY_SIZE(sampleEntries)};
     }
 
     void SampleComponentManager::RegisterSampleComponent(const SampleEntry& sample)
@@ -327,14 +327,13 @@ namespace AtomSampleViewer
 
     void SampleComponentManager::Init()
     {
-        const AZStd::vector<SampleEntry>& samples = GetSamples();
+        AZStd::array_view<SampleEntry> samples = GetSamples();
         for (const SampleEntry& sample : samples)
         {
             RegisterSampleComponent(sample);
         }
 
         m_scriptManager = AZStd::make_unique<ScriptManager>();
-
     }
 
     void SampleComponentManager::Activate()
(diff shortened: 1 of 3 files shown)


santorac:
It's about time someone did this, haha. Thanks for jumping on it!

Gem/Code/Source/SampleComponentManager.cpp
```diff
@@ -206,11 +228,80 @@ namespace AtomSampleViewer
         dependent.push_back(AZ_CRC("AzFrameworkConfigurationSystemComponentService", 0xcc49c96e)); // Ensures a scene is created for the GameEntityContext
     }
 
+    const AZStd::vector<SampleEntry>& SampleComponentManager::GetSamples()
+    {
+        static AZStd::vector<SampleEntry> sampleEntries{
```
santorac, line 233: I don't think you are allowed to make a static vector because it uses allocators that are lifetime managed. Since this function is only called once, I suggest you just construct and return a copy.

Gem/Code/Source/SampleComponentManager.cpp
```diff
@@ -241,66 +332,11 @@ namespace AtomSampleViewer
             return SampleComponentManager::IsMultiViewportSwapchainSampleSupported();
         };
```
santorac, line 333: It looks like isSupportedFunc is not used anymore. Delete.

santorac (changes requested):

Gem/Code/Source/SampleComponentManager.cpp
jeremyong-az: In this case it works because this static variable is scoped within a function and this is called after the module itself initializes. Since this vector is used twice (once to query component descriptors on module entry, once to initialize the sample component needed to build the menus), I'd rather not have to build it twice if not needed. Thoughts?

mrieggeramzn (commented):
Nice change. Always was a pain to try and find things in that big and unordered menu.

Gem/Code/Source/SampleComponentManager.cpp
mrieggeramzn: Hmm, how about the order during program termination? If the allocator is destroyed before static variable is destroyed?

jeremyong: I believe what happens at program exit is that entire virtual memory pages are returned back to the OS. That said I don't mind just changing this to make the storage completely static.

jeremyong-az: This is done now, no allocations on startup

mrieggeramzn (approved):

Gem/Code/Source/SampleComponentManager.cpp
```diff
@@ -148,44 +148,66 @@ namespace AtomSampleViewer
...
-        entry.m_sampleName = name;
-        entry.m_sampleUuid = uuid;
-        entry.m_pipelineType = SamplePipelineType::RPI;
+        SampleEntry entry = NewSample<T>(SamplePipelineType::RHI, "RHI", name, isSupportedFunction);
         entry.m_isSupportedFunc = isSupportedFunction;
         return entry;
     }
 
+    template <typename T>
+    static SampleEntry NewRPISample(const AZStd::string& name)
```
Adi-Amazon, line 188: nit / option: use macro define to create any 'type' of sample using the string / variable provided.

Adi-Amazon (approved):

Gem/Code/Source/SampleComponentManager.cpp
jeremyong-az: What's your rationale for the macro preference here?

jeremyong:
(changes still pending from @santorac )

Gem/Code/Source/SampleComponentManager.cpp
```diff
@@ -292,8 +292,9 @@ namespace AtomSampleViewer
             NewFeaturesSample<TonemappingExampleComponent>("Tonemapping"),
             NewFeaturesSample<TransparencyExampleComponent>("Transparency"),
         };
+        constexpr static size_t sampleCount = sizeof(sampleEntries) / sizeof(SampleEntry);
```
santorac, line 295: We also have the "AZ_ARRAY_SIZE" macro you could use here.

santorac (approved):

PR #309, opened 2022-01-05
Fix screenshot sample tests which rely on previous sample name convention

jeremyong-az:
A previous change modified the sample names to have a layer of nesting and this change broke ASV screenshot tests (missed because the names were referenced in lua script and not C++)


mrieggeramzn (approved):
Thanks for the super quick PR

jiaweig-amzn (approved):

PR #315, opened 2022-01-10
Fixed LockFrameTime script function.

santorac:
There were changes to the timing system recently, replacing old Cry timing code with the new AzTimer stuff. Along with these changes, t_frameTimeOverride (seconds) was repalced with a new t_simulationTickDeltaOverride (milliseconds), but LockFrameTime was not updated to convert the units accordingly.

### DIFF
--- a/Gem/Code/Source/Automation/ScriptManager.cpp
+++ b/Gem/Code/Source/Automation/ScriptManager.cpp
@@ -903,7 +903,9 @@ namespace AtomSampleViewer
         {
             AZ_DEBUG_STATIC_MEMEBER(instance, s_instance);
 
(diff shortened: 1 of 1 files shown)


Gem/Code/Source/Automation/ScriptManager.cpp
```diff
@@ -903,7 +903,9 @@ namespace AtomSampleViewer
         {
             AZ_DEBUG_STATIC_MEMEBER(instance, s_instance);
 
-            AZ::Interface<AZ::IConsole>::Get()->PerformCommand(AZStd::string::format("t_simulationTickDeltaOverride %f", seconds).c_str());
```
jeremyong-az, line 906: should we rename the variable name?

jeremyong-az (approved):

mrieggeramzn (approved):

PR #321, opened 2022-01-13
Improve shadow quality in sponza

mrieggeramzn:
CSM shadows could be made a bit nicer by tweaking the far distance. This counteracts the reduction in quality when the camera far plane was set to 1000 (from 100) in a recent change.

Before
after

After
BEFORE

Signed-off-by: mrieggeramzn <mriegger@amazon.com>


rgba16f (approved):

dmcdiarmid-ly (approved):

VickyAtAZ (approved):

Adi-Amazon (approved):

Adi-Amazon (commented):
Is there a way to detect discrepancy such as this and perhaps add a warning message so that the user / production might have a hint regarding the reason?

PR #325, opened 2022-01-25
Add function to export screenshot diffs in ASV.

hershey5045:
This is one of the pieces for the ASV pre-checkin wizard. This PR covers the following:
* When a screenshot test fails, an "Export Png Diff" button will appear in the test.
* ExportImageComparison function added to save the 1). golden image, 2). actual image, 3). and diff image.
* The idea is that users can simply drag and drop this png into their PR requests. 
* Some refactoring to regenerate a timestamp by the ScriptReporter whenever a new test script is run.

Image diff sample
imageComparison_arealighttest_polygon_double_sided_vary_rough_metal_434367 7188


jromnoa (approved):

Gem/Code/Source/Automation/ScriptReporter.cpp
```diff
@@ -619,6 +627,30 @@ namespace AtomSampleViewer
                                 ShowDiffButton("View Diff", screenshotResult.m_officialBaselineScreenshotFilePath, screenshotResult.m_screenshotFilePath);
                                 ImGui::PopID();
 
+                                if (screenshotResult.m_officialComparisonResult.m_resultCode == ImageComparisonResult::ResultCode::ThresholdExceeded && ImGui::Button("Export Png Diff"))
```
santorac, line 630: Sometimes the user might want to export the diff even if the threshold did not fail. You could do something like the "Force Show 'Update' Buttons" button above to allow users to get the export buttons.

Gem/Code/Source/Automation/ScriptReporter.cpp
```diff
@@ -619,6 +627,30 @@ namespace AtomSampleViewer
...
+                                    AzFramework::StringFunc::Path::StripExtension(screenshotFilenameWithouExtension);
+
+                                    AZStd::string timestring = GenerateTimestamp();
+
+                                    AZStd::string imageDiffFilename = "imageDiff_" +
+                                        scriptFilenameWithouExtension + "_" +
+                                        screenshotFilenameWithouExtension + "_" +
+                                        m_uniqueTimestamp +
+                                        ".png";
+                                    AzFramework::StringFunc::Path::Join(projectPath.c_str(), AZStd::string("TestResults/" + imageDiffFilename).c_str(), imageDiffPath);
```
santorac, line 649: I think this should go into the user folder, which I think would be...
```suggestion
                                    AzFramework::StringFunc::Path::Join(projectPath.c_str(), AZStd::string("user/TestResults/" + imageDiffFilename).c_str(), imageDiffPath);
```
Otherwise, I think the AP will process these images that you export.

santorac: Actually, you might also want to use separate Join calls for each folder, rather than assuming the delimiter is "/"

Gem/Code/Source/Automation/ScriptReporter.cpp
```diff
@@ -619,6 +627,30 @@ namespace AtomSampleViewer
...
+                                    ExportImageDiff(imageDiffPath.c_str(), screenshotResult);
+                                    m_messageBox.OpenPopupMessage("Image Diff Exported Successfully", AZStd::string::format("The image diff file was saved in %s", imageDiffPath.c_str()).c_str());
```
santorac, line 651: Do we have access to some platform-independent way to also have a "Open in Explorer" or "Copy path to clipboard" button?

Gem/Code/Source/Automation/ScriptReporter.cpp
```diff
@@ -1126,14 +1157,31 @@ namespace AtomSampleViewer
...
 
+    void ScriptReporter::ExportImageDiff(const char* filePath, const ScreenshotTestInfo& screenshotTestInfo)
+    {
+        using namespace AZ::Utils;
+        PngFile officialBaseline = PngFile::Load(screenshotTestInfo.m_officialBaselineScreenshotFilePath.c_str());
+        PngFile actualScreenshot = PngFile::Load(screenshotTestInfo.m_screenshotFilePath.c_str());
+
+        const size_t bufferSize = officialBaseline.GetBuffer().size();
+
+        AZStd::vector<uint8_t> diffBuffer = AZStd::vector<uint8_t>(bufferSize);
```
santorac, line 1168: Right now you are only allowing the export when the sizes match, by checking for ThresholdExceeded. This function will need some work if we ever want to allow screenshots of different size.

santorac (approved):
I left some comments but nothing blocking.

Gem/Code/Source/Automation/ScriptReporter.cpp
hershey5045: There is an ImGui function to copy text into the clipboard. Let me see if there's an open in explorer one, I think that would be ideal.

Gem/Code/Source/Automation/ScriptReporter.cpp
hershey5045: Yes that was the intent. Will there ever be a case where we need to compare screenshots of different sizes?

Gem/Code/Source/Automation/ScriptReporter.cpp
hershey5045: For now, I'll find a way to improve this experience in the wizard.

PR #334, opened 2022-02-03
Add AZ:: to RHI-namespace classes used in ScriptReporter.

hershey5045:
This was compiling for me before the changes but saw that the missing AZ namespace caused a compilation error on @HogJonny-AMZN's machine. 


jeremyong-az (approved):

Adi-Amazon (commented):
Why was this closed? it is a good fix that I was just about to suggest to you myself? :)
otherwise the compilation is broken

hershey5045:
Re-opening since this seems to be an issue present on a freshly-cloned repo. It does make sense since the AZ namespace is not available in the function

Adi-Amazon (approved):

PR #340, opened 2022-02-08
ATOM-16177 [RHI][Android] Light Culling sample crashes for device lost

jiaweig-amzn:
The initial value is too big for android and it will reach android limit. As we have a slide bar for the light number, it won't affect the ability for testing other platforms.

### DIFF
--- a/Gem/Code/Source/LightCullingExampleComponent.cpp
+++ b/Gem/Code/Source/LightCullingExampleComponent.cpp
@@ -96,7 +96,7 @@ namespace AtomSampleViewer
         m_sampleName = "LightCullingExampleComponent";
 
         // Add some initial lights to illuminate the scene
-        m_settings[(int)LightType::Point].m_numActive = 150;
+        m_settings[(int)LightType::Point].m_numActive = 10;
         m_settings[(int)LightType::Disk].m_intensity = 40.0f;
         m_settings[(int)LightType::Capsule].m_intensity = 10.0f;
     }


mrieggeramzn:
Interesting. Any idea what is going wrong with the larger number?

Gem/Code/Source/LightCullingExampleComponent.cpp
```diff
@@ -96,7 +96,7 @@ namespace AtomSampleViewer
         m_sampleName = "LightCullingExampleComponent";
 
         // Add some initial lights to illuminate the scene
-        m_settings[(int)LightType::Point].m_numActive = 150;
+        m_settings[(int)LightType::Point].m_numActive = 10;
```
moudgils, line 99: This will change the golden screenshot. you will need to update that too.

jiaweig-amzn: Oh Thanks. I'll update the screenshot as well.

jiaweig-amzn:
> Interesting. Any idea what is going wrong with the larger number?

So the crash comes from Queue::SubmitCommandBuffers at vkQueueSubmit, returning VK_ERROR_INITIALIZATION_FAILED. This is not a valid result code from vkQueueSubmit.
According to this page https://www.saschawillems.de/blog/2019/03/08/getting-a-vulkan-application-up-and-running-on-a-low-spec-device-with-buggy-drivers/
This error is a hint to timing out by the driver for processing 150 point lights.

moudgils (approved):
This change is fine as long as you also update the golden image for this sample

PR #341, opened 2022-02-09
Update AtomSampleViewer to have a new 100K Draw 10K Entity performance sample

rgba16f:
Adds a new performance submenu in the samples menu.
Removes the old RPI/HighInstanceTest from the samples menu.
Add Performance/100KEntity sample
Add Performance/100KDraw10KEntityTest sample
Updates the EntityLatticeComponent to parameterize lattice size, lattice spacing, entity size, and IBL exposure.
Add parameters to the HighInstance base class. Can now control:

1. The number of shadow casting lights ( N spot lights + 0-1 directional light)
2. Shadowmap size and filtering method.
2. Spotlight max distance, cone angles, and intensity.
3. Directional light number of cascades and intensity.
2. Lattice parameters
3. Camera positioning


rgba16f:
image

Gem/Code/Source/Performance/100KEntityExampleComponent.h
```diff
@@ -0,0 +1,36 @@
...
+
+namespace AtomSampleViewer
+{
+    /*
+        This test loads a 46x46x46 lattice of entities with randomized meshes and materials. This test is used as a simple cpu performance stress test for Atom.
+
+        The assets that are applied to the entities are chosen from the 
+        "allow-list" of models and materials.
+    */
+    class HundredKEntityExampleComponent
```
jiaweig-amzn, line 21: I guess it would be better to rename this to a number-unrelated name. What if later we want to adjust the example to be 200k or 50k? And searching becomes difficult because here already has "HundredK" and "100K".

rgba16f: What would you suggest as a name? I went with numbers because it's how we have described our cpu performance goal, 100K draws with 10K entities.

jiaweig-amzn: What do you think of directly CpuPerfomanceGoalExampleComponent?
Or I am thinking of defining a performance level (similar to what we did in screenshot comparison), and call this PerformanceLevelOneExampleComponent. So later if we want to raise the bar, we can add a level 2 and adjust the numbers in level 1. Maybe others have better idea.

rgba16f:
Successful jenkins build - https://jenkins-pipeline.agscollab.com/blue/organizations/jenkins/o3de-atom-sampleviewer-dev/detail/rgba16f%2F100KDrawsPerfTests/3/pipeline

Gem/Code/Source/EntityLatticeTestComponent.h
```diff
@@ -34,13 +36,20 @@ namespace AtomSampleViewer
         //! Returns total number of instances (width * height * depth)
         uint32_t GetInstanceCount() const;
         
+        //! Returns world space Aabb for the lattice.
```
santorac, line 39: Maybe indicate the AABB does not encompass the entity models but only their origins.

Gem/Code/Source/Performance/100KDraw10KEntityExampleComponent.h
```diff
@@ -0,0 +1,36 @@
...
+#include <Performance/HighInstanceExampleComponent.h>
+
+namespace AtomSampleViewer
+{
+    /*
+        This test loads a 46x46x46 lattice of entities with randomized meshes and materials. This test is used as a simple cpu performance stress test for Atom.
+
+        The assets that are applied to the entities are chosen from the 
+        "allow-list" of models and materials.
+    */
```
santorac, line 20: Doesn't match the official commenting style: https://github.com/o3de/sig-core/blob/main/governance/API-Ref-Guidelines-Update.md#how-should-i-format-my-code-comments

Gem/Code/Source/Performance/100KDraw10KEntityExampleComponent.h
```diff
@@ -0,0 +1,36 @@
...
+ */
+
+#pragma once
+
+#include <Performance/HighInstanceExampleComponent.h>
+
+namespace AtomSampleViewer
+{
+    /*
+        This test loads a 46x46x46 lattice of entities with randomized meshes and materials. This test is used as a simple cpu performance stress test for Atom.
```
santorac, line 16:
```suggestion
        This test loads a 22x22x22 lattice of entities with randomized meshes and materials. This test is used as a simple cpu performance stress test for Atom.
```

Gem/Code/Source/Performance/100KEntityExampleComponent.h
santorac: I'm actually okay with having the numbers in the name for now. It's very clear and it's a 2-way-door decision. BUT another approach I thought of is to make a generic "HighEntityExampleComponent" sample, instead of having separate 100KEntity and 100KDraw10KEntity samples. The lattice component provides controls for adjusting the scale, we could add controls for adjusting the lights, and then also add some "Preset" buttons to the sidebar. Then when you open the same you can adjust all the settings as you want, or you can click one of the preset buttons to pick the "100KEntity" and "100KDraw10KEntity" presets, or whatever other presets we want to add in the future.

Gem/Code/Source/Performance/100KDraw10KEntityExampleComponent.h
rgba16f: I'll update.

Gem/Code/Source/Performance/100KDraw10KEntityExampleComponent.h
rgba16f: good catch.

Gem/Code/Source/Performance/100KEntityExampleComponent.h
rgba16f: I also prefer the numbers because it's more informative to people outside Atom or even Amazon. They can understand what is being tested faster.

Gem/Code/Source/Performance/HighInstanceExampleComponent.cpp
```diff
@@ -232,10 +212,23 @@ namespace AtomSampleViewer
...
+        if (m_testParameters.m_activateDirectionalLight && m_directionalLightEnabled)
+        {
+            CreateDirectionalLight();
+        }
     }
 
     void HighInstanceTestComponent::OnAllAssetsReadyActivate()
     {
+        m_directionalLightFeatureProcessor = m_scene->GetFeatureProcessor<Render::DirectionalLightFeatureProcessorInterface>();
+        m_diskLightFeatureProcessor = m_scene->GetFeatureProcessor<Render::DiskLightFeatureProcessorInterface>();
```
santorac, line 230: Activate() already does this, is it redundant?

Gem/Code/Source/Performance/HighInstanceExampleComponent.cpp
```diff
@@ -378,4 +422,130 @@ namespace AtomSampleViewer
...
+        };
+
+        const AZ::Color& result = colors[colorIndex];
+        colorIndex = (colorIndex + 1) % colors.size();
+        return result;
+    }
+
+    AZ::Vector3 HighInstanceTestComponent::GetRandomDirection()
+    {
+        // returns a position in the range [-0.5, +0.5]^3.
```
santorac, line 481: Nitpick: this comment isn't right, you can probably just delete it

santorac (commented):
Now that I've gone through the rest of the PR and understand it better, I think it would be better to go back to having the single HighInstanceTest. It appears the only reason you are splitting this into two separate samples is because \_AutomatedPeriodicBenchmarkSuite\_.bv.lua isn't set up for any kind of parameterization of each sample, so you parameterized them by making separate samples. This goes against the original design pattern of ASV automation, where a sample should provide mechanisms for generally testing a particular feature, and specific automated test cases are achieved by scripted configuration of those samples. 

I think we can improve \_AutomatedPeriodicBenchmarkSuite\_.bv.lua to support running the same sample multipole times with different configurations using lua scripts, similar to how \_fulltestsuite\_.bv.lua can run different test scripts which potentially run the same sample.

I don't necessarily want to block the PR for this but I would like to at least have a followup to go back and improve this.

rgba16f:
> Now that I've gone through the rest of the PR and understand it better, I think it would be better to go back to having the single HighInstanceTest. It appears the only reason you are splitting this into two separate samples is because _AutomatedPeriodicBenchmarkSuite_.bv.lua isn't set up for any kind of parameterization of each sample, so you parameterized them by making separate samples. This goes against the original design pattern of ASV automation, where a sample should provide mechanisms for generally testing a particular feature, and specific automated test cases are achieved by scripted configuration of those samples.
> 
> I think we can improve _AutomatedPeriodicBenchmarkSuite_.bv.lua to support running the same sample multipole times with different configurations using lua scripts, similar to how _fulltestsuite_.bv.lua can run different test scripts which potentially run the same sample.
> 
> I don't necessarily want to block the PR for this but I would like to at least have a followup to go back and improve this.

I'm going to argue that the current ASV automation pattern is something we should move away from because it implicitly assumes that we are on a platform with mouse support. We've already found that interacting with the controls is close to impossible on mobile platforms, I suspect it would be similarly fun on consoles. On those platforms the better approach is to have a test that can be run with a command line argument.
I'd also like to argue that individual samples are more visible to non-atom team members, as they take very little understanding of ASV to run correctly.

Gem/Code/Source/EntityLatticeTestComponent.h
rgba16f: I prefer this to making the lattice AABB include the mesh AABB.

Gem/Code/Source/Performance/HighInstanceExampleComponent.cpp
rgba16f: Good call, I will try to remove.

Gem/Code/Source/Performance/HighInstanceExampleComponent.cpp
rgba16f: 👍

Gem/Code/Source/Platform/Android/EntityLatticeTestComponent_Traits_Platform.h
```diff
@@ -12,3 +12,8 @@
 #define ENTITY_LATTICE_TEST_COMPONENT_HEIGHT                    2
 #define ENTITY_LATTICE_TEST_COMPONENT_DEPTH                     2
 #define ENTITY_LATTEST_TEST_COMPONENT_MAX                       4
+#define ENTITY_LATTICE_TEST_COMPONENT_SPACING_X                 5.0f
+#define ENTITY_LATTICE_TEST_COMPONENT_SPACING_Y                 5.0f
+#define ENTITY_LATTICE_TEST_COMPONENT_SPACING_Z                 5.0f
+#define ENTITY_LATTEST_TEST_COMPONENT_SPACING_MAX               100.0f
+#define ENTITY_LATTEST_TEST_COMPONENT_ENTITY_SCALE_MAX          10.0f
```
jiaweig-amzn, line 19: Spacing and scaling don't really matter for platforms. Is there any concern that you added those?

Gem/Code/Source/Performance/HighInstanceExampleComponent.h
```diff
@@ -0,0 +1,182 @@
...
+    /*
+        This class is used as base of a set of simple cpu performance stress test for Atom.
+        This test loads a X*Y*Z lattice of entities with randomized meshes and materials and creates N shadow casting spotlights plus 0-1 Directional lights
+    */
+
+   struct HighInstanceTestParameters
+   {
+       int m_latticeSize[3] = {22, 22, 22};
+       float m_latticeSpacing[3] = {5.0f, 5.0f, 5.0f};
+       float m_entityScale = 1.0f;
```
jiaweig-amzn, line 32: We should apply here ATOMSAMPLEVIEWER_TRAIT_HIGH_INSTANCE_COUNT_TEST_COMPONENT_LATTICE_SIZE_X.
Spacing doesn't really matter, as in my other comment.

Gem/Code/Source/Performance/100KEntityExampleComponent.cpp
```diff
@@ -0,0 +1,52 @@
...
+    }
+
+
+    HundredKEntityExampleComponent::HundredKEntityExampleComponent() 
+    {
+        m_sampleName = "100KEntityTest";
+
+        for (int index = 0; index < 3; ++index)
+        {
+            m_testParameters.m_latticeSize[index] = 46;
```
jiaweig-amzn: Same as the other component. We should define another traits for this example.

jiaweig-amzn (commented):
I am OK with the component name. Some other suggestions about PAL.

Gem/Code/Source/Performance/100KDraw_10KDrawable_MultiView_ExampleComponent.cpp
```diff
@@ -0,0 +1,59 @@
...
+    }
+
+
+    _100KDraw10KDrawableExampleComponent::_100KDraw10KDrawableExampleComponent() 
+    {
+        m_sampleName = "100KDraw10KEntityTest";
+
+        for (int index = 0; index < 3; ++index)
+        {
+            m_testParameters.m_latticeSize[index] = 22;
```
jiaweig-amzn, line 38: We should define another traits for this example, at the same file as ATOMSAMPLEVIEWER_TRAIT_HIGH_INSTANCE_COUNT_TEST_COMPONENT_LATTICE_SIZE_X.
I personally prefer using vector in this case, it can be easier to assign values.

Gem/Code/Source/Performance/100KEntityExampleComponent.cpp
jiaweig-amzn: Git ate my other comment because I commented on the old file name...
See below.

Gem/Code/Source/EntityLatticeTestComponent.cpp
```diff
@@ -68,21 +70,23 @@ namespace AtomSampleViewer
...
-                        static_cast<float>(y) * distance.GetY(),
-                        static_cast<float>(z) * distance.GetZ());
+                        static_cast<float>(x) * m_spacingX,
+                        static_cast<float>(y) * m_spacingY,
+                        static_cast<float>(z) * m_spacingZ);
 
                     transform.SetTranslation(position);
+                    transform.SetUniformScale(m_entityScale);
                     CreateLatticeInstance(transform);
+                    m_worldAabb.AddPoint(transform.GetTranslation());
```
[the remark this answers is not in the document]
rgba16f: That would be hard to achieve during activation and would result in the aabb changing as assets loaded. I prefer the other approach of documenting that the AABB just includes the lattice positions.

Gem/Code/Source/Platform/Android/EntityLatticeTestComponent_Traits_Platform.h
rgba16f: In this case I was following the lattice size example. I can remove the spacing and scale Pal-ification if necessary.

Gem/Code/Source/Platform/Windows/HighInstanceTestComponent_Traits_Platform.h
```diff
@@ -6,6 +6,6 @@
  *
  */
 
-#define ATOMSAMPLEVIEWER_TRAIT_HIGH_INSTANCE_COUNT_TEST_COMPONENT_LATTICE_SIZE_X       22
-#define ATOMSAMPLEVIEWER_TRAIT_HIGH_INSTANCE_COUNT_TEST_COMPONENT_LATTICE_SIZE_Y       22
-#define ATOMSAMPLEVIEWER_TRAIT_HIGH_INSTANCE_COUNT_TEST_COMPONENT_LATTICE_SIZE_Z       22
+#define ATOMSAMPLEVIEWER_TRAIT_HIGH_INSTANCE_COUNT_TEST_COMPONENT_LATTICE_SIZE_X       100
+#define ATOMSAMPLEVIEWER_TRAIT_HIGH_INSTANCE_COUNT_TEST_COMPONENT_LATTICE_SIZE_Y       100
+#define ATOMSAMPLEVIEWER_TRAIT_HIGH_INSTANCE_COUNT_TEST_COMPONENT_LATTICE_SIZE_Z       10
```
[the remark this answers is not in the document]
rgba16f: My plan for android and other platforms is that we would have specific performance tests for them. The current tests are for the 2022 goals on windows desktops. With that in mind I don't think palifying the test parameters is useful.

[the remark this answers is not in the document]
jiaweig-amzn: I agree that this test is not suitable for android. Palifying is to avoid having android frozen in this sample or even crash for any resources out of limit. In fact, some other samples have the same issue.
I think the eventual scenario is to have android running a different set of samples, but this is not implemented. For now, we can just use the macros.

rgba16f:
Green build with the latest changes - https://jenkins-pipeline.agscollab.com/job/o3de-atom-sampleviewer-dev/job/rgba16f%2F100KDrawsPerfTests/9/

jiaweig-amzn (approved):

Gem/Code/Source/EntityLatticeTestComponent.h
rgba16f: Comment was updated.

rgba16f:
> 

I disagree that the performance tests should go back to a single test approach for several reasons. 

1. All of the existing ASV samples are correctness tests, and so need to work on all platforms. The performance tests are known workloads and so are expected work poorly or to not work on some platforms. The ability to have a workload that fails on some platforms is a desirable characteristic as it means we can point vendors to a simple test that fails on their platform. 
2. As we work on more platforms it will be useful to have a ladder of known workload performance tests, so we can see how Atom scales across various parameters. Plus so we can give performance goal recommendations to external teams.
3. Having multiple tests makes it easy to test performance on platforms where we don't have a ui running yet (AR/VR) or where imgui is hard to use (small screen mobile, other platforms without mouse support)

santorac (approved):

Gem/Code/Source/EntityLatticeTestComponent.cpp
```diff
@@ -94,11 +98,33 @@ namespace AtomSampleViewer
...
+    void EntityLatticeTestComponent::SetLatticeSpacing( float spaceX, float spaceY, float spaceZ)
+    {
+        m_spacingX = AZ::GetClamp<float>(spaceX, 0.0f, s_spacingMax);
+        m_spacingY = AZ::GetClamp<float>(spaceY, 0.0f, s_spacingMax);
+        m_spacingZ = AZ::GetClamp<float>(spaceZ, 0.0f, s_spacingMax);
+    }
+
+    void EntityLatticeTestComponent::SetLatticeEntityScale(float scale)
+    {
+        m_entityScale = AZ::GetClamp<float>(scale, 0.0f, s_entityScaleMax);
```
antonmic, line 122: Similarly, do we want the use case where the scale is 0? Seems like inviting user error.
If someone wants a negative scale, the scale will be clamped to 0 and nothing will render, correct?

Gem/Code/Source/Performance/100KDraw_10KDrawable_MultiView_ExampleComponent.cpp
```diff
@@ -0,0 +1,60 @@
...
+
+        m_testParameters.m_activateDirectionalLight = true;
+        m_testParameters.m_directionalLightIntensity = 5.0f;
+
+        m_testParameters.m_cameraPosition[0] = -81.0f;
+        m_testParameters.m_cameraPosition[1] = 31.0f;
+        m_testParameters.m_cameraPosition[2] = 32.0f;
+        m_testParameters.m_cameraHeadingDeg = -90.0f;
+        m_testParameters.m_cameraPitchDeg = 0.0f;
+        m_testParameters.m_iblExposure = -10.0f;
```
antonmic, line 56: Minor nit: If you wrap this in a "InitializeDefaultValues()" function, the magic numbers would appear a bit less magical and you could more easily add a "reset()" option later to reset camera position etc.

antonmic (approved):
Looks good, ship it!

PR #342, opened 2022-02-10
Added MimimalMultilayerPbr Material Type For Testing Nested Property Groups

santorac:
Added a MinimalMultilayerPBR material type and screenshot test for regression testing the nested material property groups feature. This is only temporary until the StandardMultilayerPBR material type can be updated and refactored to utilize these new features, at which point MinimalMultilayerPBR can be removed.

See https://github.com/o3de/o3de/pull/7537


gadams3 (approved):

Materials/Types/MinimalMultilayerPBR_LayerProperties.json
```diff
@@ -0,0 +1,75 @@
...
+            "max": 1.0,
+            "connection": {
+                "type": "ShaderInput",
+                "name": "roughness"
+            }
+        }
+    ],
+    "propertyGroups": [
+        {
+            "name": "baseColor",
```
amzn-tommy, line 32: Can we nest imports, so that this imports the baseColor properties from another file (that is in turn, also used by StandardPBR)? If so, it might be worth doing that here to validate that importing a file which in turn imports another file is something that works.

amzn-tommy (approved):

HogJonny-AMZN (approved):

Materials/Types/MinimalMultilayerPBR_LayerProperties.json
santorac: Yes we can nest them. I have a unit tests to verify this. I'll exercise the nesting in actual data when I refactor our main material types. My goal for MinimalMultilayerPbr was more around testing multi-level nesting and name contexts in Material Editor and Material Component UI and .material file saving.

PR #344, opened 2022-02-17
Update Framerate Histogram to be a Frame Time histogram

rgba16f:
present average, min, max frame time in milliseconds

### DIFF
--- a/Gem/Code/Source/Utils/ImGuiHistogramQueue.cpp
+++ b/Gem/Code/Source/Utils/ImGuiHistogramQueue.cpp
@@ -27,17 +27,23 @@ namespace AtomSampleViewer
         m_averageLog.reserve(m_maxSamples);
     }
 
-    float ImGuiHistogramQueue::CalculateAverage(AZStd::size_t maxSampleCount)
+    float ImGuiHistogramQueue::UpdateDisplayedValues(AZStd::size_t maxSampleCount, float& minValue, float& maxValue)
     {
         size_t sampleCount = AZStd::min<size_t>(maxSampleCount, m_valueLog.size());
-
-        float average = 0.0f;
-        for (size_t i = 0; i < sampleCount; ++i)
+        float average = 0.f;
+        if (sampleCount > 0)
         {
-            average += m_valueLog.at(i);
+            average = m_valueLog.at(0);
+            minValue = maxValue = average;
+            for (size_t i = 1; i < sampleCount; ++i)
+            {
+                float curValue = m_valueLog.at(i);
+                average += curValue;
+                minValue = minValue > curValue ? curValue : minValue;
+                maxValue = maxValue < curValue ? curValue : maxValue;
+            }
+            average /= sampleCount;
         }
-        average /= sampleCount;
-
         return average;
     }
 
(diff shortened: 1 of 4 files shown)


Gem/Code/Source/Utils/ImGuiHistogramQueue.cpp
```diff
@@ -1,105 +1,112 @@
...
+        m_averageLog.reserve(m_maxSamples);
+    }
+
+    float ImGuiHistogramQueue::UpdateDisplayedValues(AZStd::size_t maxSampleCount, float& minValue, float& maxValue)
+    {
+        size_t sampleCount = AZStd::min<size_t>(maxSampleCount, m_valueLog.size());
+        float average = 0.f;
+        if (sampleCount > 0)
+        {
+            average = m_valueLog.at(0);
```
jeremyong-az, line 36: In the STL anyways, the distinction between `at` and `operation[]` is that the former throws an exception for out of bounds access. Is that intended here?

Gem/Code/Source/Utils/ImGuiHistogramQueue.cpp
```diff
@@ -1,105 +1,112 @@
...
+    {
+        AZ_Assert(m_maxSamples >= m_runningAverageSamples, "maxSamples must be larger");
+
+        m_valueLog.reserve(m_maxSamples);
+        m_averageLog.reserve(m_maxSamples);
+    }
+
+    float ImGuiHistogramQueue::UpdateDisplayedValues(AZStd::size_t maxSampleCount, float& minValue, float& maxValue)
+    {
+        size_t sampleCount = AZStd::min<size_t>(maxSampleCount, m_valueLog.size());
```
jeremyong-az, line 32: I think as of C++17, the template argument `size_t` here should be deducible.

Gem/Code/Source/Utils/ImGuiHistogramQueue.cpp
```diff
@@ -1,105 +1,112 @@
...
+    void ImGuiHistogramQueue::PushValue(float value)
+    {
+        m_samplesSinceLastDisplayUpdate++;
+
+        // Update the log of all values
+        if (m_valueLog.size() == m_maxSamples)
+        {
+            m_valueLog.pop_back();
+        }
+        m_valueLog.insert(m_valueLog.begin(), value);
```
jeremyong-az, line 59: This is an interesting way to implement a ring buffer (I see that it's carry-over from before, but suggest we actually just use/provide a ring-buffer at some point)

jeremyong-az (approved):
Was there a reason every line in the files needed to change?

wintermute-motherbrain (approved):

rgba16f:
> Was there a reason every line in the files needed to change?

It wasn't intentional some tool of mine decided to modify every line ending in the file.

Gem/Code/Source/Utils/ImGuiHistogramQueue.cpp
rgba16f: The existing code was using `at`, not having a strong reason to change it I left it as is.

Gem/Code/Source/Utils/ImGuiHistogramQueue.cpp
jeremyong-az: FYI @rgba16f we actually do have an AZStd::ring_buffer if you want to use that instead

rgba16f:
Successful build - https://jenkins-pipeline.agscollab.com/view/Projects/job/o3de-atom-sampleviewer-dev/job/rgba16f%252FHistogramFpsToMs/2/
Successful ASV run - 
image

PR #349, opened 2022-02-21
Increase thresholds for a couple screenshot tests

santorac:
Increase thresholds for a couple screenshot tests that were failing on Vulkan with benign differences.

MinimalMultilayerPbrExample

**Note for the depth-stencil tests,** there is a giant difference in the alpha channel, which is not actually checked as part of these results. So ignore that difference, focus on the minor differences in the rest of the image. Here is another PR to address the alpha comparison issue https://github.com/o3de/o3de/pull/7743

depthStencilMs
depthStencilResolve


moudgils (approved):

smurly (approved):

VickyAtAZ (approved):

VickyAtAZ:
The depth stencil one is actually a real bug which I am working on (which only happens on Vulkan).

PR #351, opened 2022-02-23
Make ProceduralSkinnedMeshUtils use model buffers directly for skinned meshes

amzn-tommy:
-Updated the SkinnedMesh sample to work with the new interface, which just takes a model asset instead of manually setting a bunch of individual buffers.
-Updated ProceduralSkinnedMesh to pack joint ids
-Updated ProceduralSkinnedMeshUtils to produce a model with an arbitrary number of sub-meshes. To keep it simple, it does so by duplicating the entire ProceduralSkinnedMesh and offsetting it along the y-axis.

New Sub-mesh functionality:
asv_skinnedmeshsample

ASV Test results:
1 failed test due to a known issue with spotlight shadow banding
Capture


amzn-tommy:
Re-opened with a new branch in https://github.com/o3de/o3de-atom-sampleviewer/pull/356 without the DCO issues

PR #356, opened 2022-03-03
Make ProceduralSkinnedMeshUtils use model buffers directly for skinned meshes

amzn-tommy:
-Updated the SkinnedMesh sample to work with the new interface, which just takes a model asset instead of manually setting a bunch of individual buffers.
-Updated ProceduralSkinnedMesh to pack joint ids
-Updated ProceduralSkinnedMeshUtils to produce a model with an arbitrary number of sub-meshes. To keep it simple, it does so by duplicating the entire ProceduralSkinnedMesh and offsetting it along the y-axis.

New Sub-mesh functionality:
asv_skinnedmeshsample

ASV Test results:
1 failed test due to a known issue with spotlight shadow banding
Capture


Gem/Code/Source/ProceduralSkinnedMesh.cpp
```diff
@@ -143,9 +161,31 @@ namespace AtomSampleViewer
             m_bitangents[vertexIndex][1] = 0.0f;
             m_bitangents[vertexIndex][2] = -1.0f;
 
-            // Blend indices/weights are the same for each vertex in the segment
-            m_blendIndices[vertexIndex] = m_segmentBlendIndices[segmentIndex];
-            m_blendWeights[vertexIndex] = m_segmentBlendWeights[segmentIndex];
+            for (size_t i = 0; i < m_influencesPerVertex; ++i)
+            {
+                // m_blendIndices has two id's packed into a single uint32
+                size_t packedIndex = vertexIndex * m_influencesPerVertex / 2 + i / 2;
```
Adi-Amazon, line 167: Why not simplify the scheme by using upper and lower 16 bits? simple and fast.

Adi-Amazon, line 167: super nit: put ( expression ) / 2 and you save one division.

Gem/Code/Source/ProceduralSkinnedMesh.cpp
```diff
@@ -285,8 +326,14 @@ namespace AtomSampleViewer
                 }
             }
 
-            m_segmentBlendIndices[segmentIndex] = currentBlendIndices;
-            m_segmentBlendWeights[segmentIndex] = currentBlendWeights;
+            // Now copy the resulting influences into the larger buffer
+            for (size_t i = 0; i < m_influencesPerVertex; ++i)
+            {
+                size_t destinationIndex = segmentIndex * m_influencesPerVertex + i;
```
Adi-Amazon, line 332: nit: you can save this multiplication by putting it before the inner for block

Gem/Code/Source/ProceduralSkinnedMeshUtils.cpp
```diff
@@ -62,12 +98,52 @@ namespace AtomSampleViewer
...
+        for (uint32_t subMeshIndex = 1; subMeshIndex < submeshCount; ++subMeshIndex)
+        {
+            for (uint32_t i = 0; i < verticesPerSubmesh; ++i)
+            {
+                proceduralMesh.m_positions[subMeshIndex*verticesPerSubmesh + i][1] +=
+                    aznumeric_cast<float>(subMeshIndex) * proceduralMesh.GetSubMeshYOffset();
+            }
+        }
+
+        auto indexBuffer = CreateTypedBufferAsset(proceduralMesh.m_indices.data(), proceduralMesh.m_indices.size(), AZ::RHI::Format::R32_FLOAT);
```
Adi-Amazon, line 139: Questions:
1. Why is the index buffer format R32_Float? it shouldn't matter but isn't it supposed to be R32_UINT?
2. What happens if the original format is 16 bits, how is that supported (probably not here where it's only the buffer creation) with the right space required?
3. Isn't there a way to simply take this from the original mesh based on its buffer descriptor?

Gem/Code/Source/ProceduralSkinnedMeshUtils.cpp
```diff
@@ -81,28 +157,59 @@ namespace AtomSampleViewer
...
+            uint32_t elementCount = verticesPerSubmesh;
+            uint32_t elementOffset = verticesPerSubmesh * submeshIndex;
+
+            // Include any truncated vertices if this is the last mesh
+            if (submeshIndex == submeshCount - 1)
+            {
+                elementCount += totalVertices % verticesPerSubmesh;
+            }
+
+            modelLodCreator.AddMeshStreamBuffer(RHI::ShaderSemantic{ "POSITION" }, AZ::Name(), AZ::RPI::BufferAssetView{ positionBuffer, CreateSubmeshBufferViewDescriptor(positionBuffer, elementCount, elementOffset) });
```
Adi-Amazon, line 193: nit: We are assuming here (as we do in other places) that first 'layer' will always have position, normal, UVs, tangent and bi-tangent. 
Should we not check for that and work based on the existing data? Even if this is the current state, working based on what exists can help generalizing it for the future.

Adi-Amazon (approved):
No blockers but some points to think about.

Gem/Code/Source/ProceduralSkinnedMeshUtils.h
```diff
@@ -28,7 +28,7 @@ namespace AtomSampleViewer
 {
     class ProceduralSkinnedMesh;
 
-    AZ::Data::Instance<AZ::RPI::Model> CreateModelFromProceduralSkinnedMesh(const ProceduralSkinnedMesh& proceduralMesh);
-    AZStd::intrusive_ptr<AZ::Render::SkinnedMeshInputBuffers> CreateSkinnedMeshInputBuffersFromProceduralSkinnedMesh(const ProceduralSkinnedMesh& proceduralMesh);
+    AZ::Data::Instance<AZ::RPI::Model> CreateModelFromProceduralSkinnedMesh(ProceduralSkinnedMesh& proceduralMesh);
```
mrieggeramzn, line 31: Feel like these functions deserve a comment indicating what happens to the input parameters.

mrieggeramzn (approved):

PR #363, opened 2022-03-09
Adding readback sample

Bindless-Chicken:
This commit introduce the readback sample. It is designed to render a pattern into a texture that is then readback to host memory and reuploaded to the GPU to be rendered.
This is only the shell of the proper test that will be updated as soon as the changes supporting subresource readback are added.

This was tested using the full test-suite on PC Windows (DX12 and Vulkan).

### DIFF
--- a/Gem/Code/Source/ReadbackExampleComponent.cpp
+++ b/Gem/Code/Source/ReadbackExampleComponent.cpp
@@ -274,7 +274,7 @@ namespace AtomSampleViewer
         m_fillerPass->ReadbackAttachment(m_readback, AZ::Name("Output"));
     }
 
-    void ReadbackExampleComponent::ReadbackCallback([[maybe_unused]] const AZ::RPI::AttachmentReadback::ReadbackResult& result)
+    void ReadbackExampleComponent::ReadbackCallback(const AZ::RPI::AttachmentReadback::ReadbackResult& result)
     {
         const AZ::RHI::ImageSubresourceRange range(0, 0, 0, 0);
         AZ::RHI::ImageSubresourceLayoutPlaced layout;
@@ -299,8 +299,7 @@ namespace AtomSampleViewer
         updateRequest.m_sourceSubresourceLayout = layout;
         updateRequest.m_sourceData = m_resultData->begin();
         updateRequest.m_imageSubresourcePixelOffset = AZ::RHI::Origin(0, 0, 0);
-        AZ::Data::Instance<AZ::RPI::AttachmentImagePool> pool = AZ::RPI::ImageSystemInterface::Get()->GetSystemAttachmentPool();
-        pool->GetRHIPool()->UpdateImageContents(updateRequest);
+        m_previewImage->UpdateImageContents(updateRequest);
     }
 
     void ReadbackExampleComponent::DrawSidebar()
--- a/Gem/Code/Source/ReadbackExampleComponent.cpp
+++ b/Gem/Code/Source/ReadbackExampleComponent.cpp
@@ -1,0 +1,343 @@
+/*
+ * Copyright (c) Contributors to the Open 3D Engine Project.
+ * For complete copyright and license terms please see the LICENSE at the root of this distribution.
+ *
+ * SPDX-License-Identifier: Apache-2.0 OR MIT
+ *
+ */
+
+#include <ReadbackExampleComponent.h>
+
+#include <Atom/RPI.Public/Scene.h>
+#include <Atom/RPI.Public/RenderPipeline.h>
+#include <Atom/RPI.Public/Pass/FullscreenTrianglePass.h>
+#include <Atom/RPI.Public/Image/ImageSystemInterface.h>
+#include <Atom/RPI.Public/Image/AttachmentImagePool.h>
+#include <Atom/RPI.Public/RPIUtils.h>
+
+#include <Atom/RPI.Reflect/Asset/AssetUtils.h>
+#include <Atom/RPI.Reflect/Pass/FullscreenTrianglePassData.h>
+
+#include <Automation/ScriptableImGui.h>
+
+namespace AtomSampleViewer
+{
+    static const char* s_readbackPipelineTemplate = "ReadbackPipelineTemplate";
+    static const char* s_fillerPassTemplate = "ReadbackFillerPassTemplate";
+    static const char* s_previewPassTemplate = "ReadbackPreviewPassTemplate";
+
+    static const char* s_fillerShaderPath = "Shaders/Readback/Filler.azshader";
+    static const char* s_previewShaderPath = "Shaders/Readback/Preview.azshader";
+
+    static const char* s_readbackImageName = "ReadbackImage";
+    static const char* s_previewImageName = "PreviewImage";
+
+    void ReadbackExampleComponent::Reflect(AZ::ReflectContext* context)
+    {
+        if (AZ::SerializeContext* serializeContext = azrtti_cast<AZ::SerializeContext*>(context))
+        {
+            serializeContext->Class<ReadbackExampleComponent, AZ::Component>()->Version(0);
+        }
+    }
+
+    ReadbackExampleComponent::ReadbackExampleComponent()
+    {
+    }
+
+    void ReadbackExampleComponent::Activate()
+    {
+        AZ::TickBus::Handler::BusConnect();
+        AZ::Render::Bootstrap::DefaultWindowNotificationBus::Handler::BusConnect();
+
+        ActivatePipeline();
+        CreatePasses();
+
+        m_imguiSidebar.Activate();
+    }
+
+    void ReadbackExampleComponent::Deactivate()
+    {
+        m_imguiSidebar.Deactivate();
+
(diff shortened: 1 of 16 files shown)


Gem/Code/Source/ReadbackExampleComponent.cpp
```diff
@@ -0,0 +1,343 @@
...
+    void ReadbackExampleComponent::ActivatePipeline()
+    {
+        // Create the pipeline
+        CreatePipeline();
+
+        // Setup the pipeline
+        m_originalPipeline = m_scene->GetDefaultRenderPipeline();
+        m_scene->AddRenderPipeline(m_readbackPipeline);
+        m_scene->RemoveRenderPipeline(m_originalPipeline->GetId());
+        m_scene->SetDefaultRenderPipeline(m_originalPipeline->GetId());
```
rgba16f, line 110: Should that last line be setting the readback pipeline?

Gem/Code/Source/ReadbackExampleComponent.cpp
```diff
@@ -0,0 +1,343 @@
...
+        AZ::Data::Instance<AZ::RPI::AttachmentImagePool> pool = AZ::RPI::ImageSystemInterface::Get()->GetSystemAttachmentPool();
+        pool->GetRHIPool()->UpdateImageContents(updateRequest);
+    }
+
+    void ReadbackExampleComponent::DrawSidebar()
+    {
+        if (m_imguiSidebar.Begin())
+        {
+            ImGui::Text("Readback resource dimensions:");
+            if (ScriptableImGui::SliderInt("Width", reinterpret_cast<int*>(& m_resourceWidth), 1, 2048))
```
moudgils, line 311: - Does static_cast not work here?
- Remove the space after &

Gem/Code/Source/ReadbackExampleComponent.cpp
```diff
@@ -0,0 +1,343 @@
...
+            {
+                PassesChanged();
+            }
+            if (ScriptableImGui::SliderInt("Height", reinterpret_cast<int*>(&m_resourceHeight), 1, 2048))
```
moudgils, line 315: static_cast?

Gem/Code/Source/ReadbackExampleComponent.h
```diff
@@ -0,0 +1,111 @@
...
+#include <Atom/Bootstrap/DefaultWindowBus.h>
+
+#include <Atom/RPI.Public/Pass/AttachmentReadback.h>
+
+#include <Utils/ImGuiSidebar.h>
+#include <Atom/Feature/ImGui/ImGuiUtils.h>
+
+namespace AtomSampleViewer
+{
+    /*
```
moudgils, line 23: Comment Style guidelines -> Use //! instead of /*

moudgils (approved):
Looks good with some minor comments.

Gem/Code/Source/ReadbackExampleComponent.cpp
```diff
@@ -0,0 +1,343 @@
...
+        AZ::Render::Bootstrap::DefaultWindowNotificationBus::Handler::BusDisconnect();
+        AZ::TickBus::Handler::BusDisconnect();
+    }
+
+    void ReadbackExampleComponent::OnTick([[maybe_unused]] float deltaTime, [[maybe_unused]] AZ::ScriptTimePoint scriptTime)
+    {
+        // Readback was completed, we need to update the preview image
+        if (m_textureNeedsUpdate)
+        {
+            UploadRedbackResult();
```
jeremyong-az, line 74: Nit: `UploadReadbackResult`

Shaders/Readback/Filler.shader
```diff
@@ -0,0 +1,34 @@
+{
+    "Source" : "Filler.azsl",
+
+    "DepthStencilState" : { 
+        "Depth" : { 
+            "Enable" : false
```
jeremyong-az, line 6: Is depth needed for this sample?

jeremyong-az (approved):

jeremyong-az:
Just a couple comments, thanks for the sample!

Gem/Code/Source/ReadbackExampleComponent.cpp
Bindless-Chicken: Good catch. I guess in that case it didn't matter because the original pipeline was removed and the readback pipeline was made default. Changed in any case.

Gem/Code/Source/ReadbackExampleComponent.cpp
Bindless-Chicken: I don't believe so, `int*` and `uint32_t*` are unrelated types.

Shaders/Readback/Filler.shader
Bindless-Chicken: Nope, it's really just rendering a uv pattern on a full screen triangle.

Bindless-Chicken:
I ran the tests (this one and full suite) on Vulkan they all go through with no screenshot errors, but I get an absurd amount of validation layer errors in every sample.
> SPIR-V module not valid: Structure id 5 decorated as BufferBlock for variable in Uniform storage class must follow relaxed storage buffer layout rules: member 0 contains an array with stride 164 not satisfying alignment to 16

Or

>vertex shader writes to output location 0.0 which is not consumed by fragment shader

When the pipeline are changed. I don't believe this is related to this particular change.

Gem/Code/Source/ReadbackExampleComponent.cpp
```diff
@@ -0,0 +1,343 @@
...
+        if (!m_readback)
+        {
+            m_readback = AZStd::make_shared<AZ::RPI::AttachmentReadback>(AZ::RHI::ScopeId{ "RenderTargetCapture" });
+            m_readback->SetCallback(AZStd::bind(&ReadbackExampleComponent::ReadbackCallback, this, AZStd::placeholders::_1));
+        }
+
+        m_fillerPass->ReadbackAttachment(m_readback, AZ::Name("Output"));
+    }
+
+    void ReadbackExampleComponent::ReadbackCallback([[maybe_unused]] const AZ::RPI::AttachmentReadback::ReadbackResult& result)
```
VickyAtAZ, line 277: nit: result is used in the function

Gem/Code/Source/ReadbackExampleComponent.cpp
```diff
@@ -0,0 +1,343 @@
...
+        const AZ::RHI::ImageSubresourceRange range(0, 0, 0, 0);
+        AZ::RHI::ImageSubresourceLayoutPlaced layout;
+        m_previewImage->GetRHIImage()->GetSubresourceLayouts(range, &layout, nullptr);
+        AZ::RHI::ImageUpdateRequest updateRequest;
+        updateRequest.m_image = m_previewImage->GetRHIImage();
+        updateRequest.m_sourceSubresourceLayout = layout;
+        updateRequest.m_sourceData = m_resultData->begin();
+        updateRequest.m_imageSubresourcePixelOffset = AZ::RHI::Origin(0, 0, 0);
+        AZ::Data::Instance<AZ::RPI::AttachmentImagePool> pool = AZ::RPI::ImageSystemInterface::Get()->GetSystemAttachmentPool();
+        pool->GetRHIPool()->UpdateImageContents(updateRequest);
```
VickyAtAZ, line 303: Optional: you can use AttachmentImage::UpdateImageContents here to upload new data.

VickyAtAZ (approved):

VickyAtAZ:
> satisfying alignment to 16

Yes. That is a known issue. I think Doug will fix it.

moudgils:
> I ran the tests (this one and full suite) on Vulkan they all go through with no screenshot errors, but I get an absurd amount of validation layer errors in every sample.
> 
> > SPIR-V module not valid: Structure id 5 decorated as BufferBlock for variable in Uniform storage class must follow relaxed storage buffer layout rules: member 0 contains an array with stride 164 not satisfying alignment to 16
> 
> Or
> 
> > vertex shader writes to output location 0.0 which is not consumed by fragment shader
> 
> When the pipeline are changed. I don't believe this is related to this particular change.

You can run profile build as the valdiations are only enabled for Debug builds. Those seem like non-fatal issues.

Bindless-Chicken:
> > I ran the tests (this one and full suite) on Vulkan they all go through with no screenshot errors, but I get an absurd amount of validation layer errors in every sample.
> > > SPIR-V module not valid: Structure id 5 decorated as BufferBlock for variable in Uniform storage class must follow relaxed storage buffer layout rules: member 0 contains an array with stride 164 not satisfying alignment to 16
> > 
> > 
> > Or
> > > vertex shader writes to output location 0.0 which is not consumed by fragment shader
> > 
> > 
> > When the pipeline are changed. I don't believe this is related to this particular change.
> 
> You can run profile build as the valdiations are only enabled for Debug builds. Those seem like non-fatal issues.

Yep, tests runs fine otherwise.

rgba16f:
It seems like this change causes a Mac build failure.
https://github.com/o3de/o3de/issues/8356

PR #364, opened 2022-03-09
Fix code for linux.

hershey5045:
Testing:
* Compiles successfully on linux

### DIFF
--- a/Gem/Code/Source/Automation/ScriptManager.cpp
+++ b/Gem/Code/Source/Automation/ScriptManager.cpp
@@ -9,7 +9,7 @@
 #include <Automation/ScriptManager.h>
 #include <Automation/ScriptableImGui.h>
 #include <SampleComponentManagerBus.h>
-
+#include <imgui/imgui_internal.h>
 #include <Atom/RPI.Reflect/Asset/AssetUtils.h>
 
 #include <Atom/RPI.Public/RPISystemInterface.h>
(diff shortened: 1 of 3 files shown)


gadams3 (approved):

Gem/Code/Source/Automation/ScriptManager.cpp
```diff
@@ -839,12 +839,12 @@ namespace AtomSampleViewer
         }
 
         ImGui::Text("RHI API: %s", AZ::RHI::Factory::Get().GetName().GetCStr());
-        ImGui::Text("Test Script Count: %d", scriptReports.size());
+        ImGui::Text("Test Script Count: %lu", scriptReports.size());
```
amzn-tommy, line 842: Assuming .size() returns a size_t, shouldn't this be %zu?

amzn-tommy (commented):
I think this is still

Gem/Code/Source/Automation/ScriptManager.cpp
```diff
@@ -807,7 +807,7 @@ namespace AtomSampleViewer
         }
         ImGui::SameLine();
         ImGui::Text(
-            "%d/%d", AZStd::distance(m_wizardSettings.m_reportsOrderedByThresholdToInspect.begin(), m_wizardSettings.m_reportIterator) + 1,
+            "%ld/%lu", AZStd::distance(m_wizardSettings.m_reportsOrderedByThresholdToInspect.begin(), m_wizardSettings.m_reportIterator) + 1,
```
amzn-tommy, line 810: I don't know what the correct conversion specifier to use for 'difference_type' is, and didn't see anything obvious at the top of a brief goole search, but if there is no clear correct option for all platforms, it might be worth casting the result of AZStd::distance to a size_t, int, etc. and then use the correct conversion specifier for that type in printf.

hershey5045: Thanks, these should be fixed now.

amzn-tommy (approved):

PR #367, opened 2022-03-15
Add settings to sort script reporter screenshots by threshold

hershey5045:
Notes:
* Script reporter dialog now shows the screenshot tests sorted in a flat list by default
image

* There is a button to display screenshot results as it was done previously
* PR might look messy since the loop on ScriptReports was tabbed, but basically:
     * the loop for ScreenshotTestInfos was refactored into a separate function.
     * if m_showReportsSortedByThreshold is false, then the previous way of displaying screenshot results will be done, otherwise the new way will take effect.

### DIFF
--- a/Gem/Code/Source/Automation/ScriptReporter.cpp
+++ b/Gem/Code/Source/Automation/ScriptReporter.cpp
@@ -147,9 +147,11 @@ namespace AtomSampleViewer
     void ScriptReporter::Reset()
     {
         m_scriptReports.clear();
+        m_descendingThresholdReports.clear();
         m_currentScriptIndexStack.clear();
         m_invalidationMessage.clear();
-        m_uniqueTimestamp = GenerateTimestamp();     
+        m_uniqueTimestamp = GenerateTimestamp();
+
     }
 
     void ScriptReporter::SetInvalidationMessage(const AZStd::string& message)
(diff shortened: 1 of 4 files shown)


Gem/Code/Source/Automation/ScriptReporter.cpp
```diff
@@ -685,6 +577,154 @@ namespace AtomSampleViewer
...
+            ImGui::Unindent();
+
+            ImGui::Spacing();
+
+            HighlightTextIf(localBaselineWarning, m_highlightSettings.m_highlightWarning);
+
+            ImGui::Text("Local Baseline:    %s", screenshotResult.m_localBaselineScreenshotFilePath.c_str());
+
+            // Local Baseline Result
+            ImGui::Indent();
```
mrieggeramzn, line 687: Would recommend extracting out functions into something like ShowLocalBaselineResult() for instance.

Gem/Code/Source/Automation/ScriptReporter.cpp
```diff
@@ -685,6 +577,154 @@ namespace AtomSampleViewer
...
+            ImGui::Text("Screenshot:        %s", screenshotResult.m_screenshotFilePath.c_str());
+
+            ImGui::Spacing();
+
+            HighlightTextIf(!screenshotPassed, m_highlightSettings.m_highlightFailed);
+
+            ImGui::Text("Official Baseline: %s", screenshotResult.m_officialBaselineScreenshotFilePath.c_str());
+
+            // Official Baseline Result
+            ImGui::Indent();
```
mrieggeramzn, line 610: Function getting pretty big. Would recommend extracting out a subfunction called, say, ShowOfficalBaselineResult().

mrieggeramzn (approved):

Gem/Code/Source/Automation/ScriptReporter.cpp
```diff
@@ -419,255 +421,145 @@ namespace AtomSampleViewer
...
+                            const bool screenshotPassed = screenshotResult.m_officialComparisonResult.m_resultCode == ImageComparisonResult::ResultCode::Pass;
+                            const bool localBaselineWarning = screenshotResult.m_localComparisonResult.m_resultCode != ImageComparisonResult::ResultCode::Pass;
 
-                            ImGui::Text("Official Baseline: %s", screenshotResult.m_officialBaselineScreenshotFilePath.c_str());
+                            AZStd::string fileName;
+                            AzFramework::StringFunc::Path::GetFullFileName(screenshotResult.m_screenshotFilePath.c_str(), fileName);
 
-                            // Official Baseline Result
-                            ImGui::Indent();
+                            AZStd::string headerSummary;
```
jiaweig-amzn, line 517: In this case, I think using string stream would be a better choice, since it is appending and constructing string all the time.

Gem/Code/Source/Automation/ScriptReporter.cpp
```diff
@@ -419,255 +421,145 @@ namespace AtomSampleViewer
...
-                    if (showAll || (showWarnings && scriptReport.m_generalWarningCount > 0))
-                    {
-                        ImGui::Text("Warnings: %u %s", scriptReport.m_generalWarningCount, SeeConsole(scriptReport.m_generalWarningCount, "Trace::Warning").c_str());
-                    }
+                    // Skip if tests passed without warnings and we don't want to show successes
+                    bool skipReport = (scriptPassed && !scriptHasWarnings && !m_showAll);
 
-                    ResetTextHighlight();
+                    // Skip if we only have warnings only and we don't want to show warnings
+                    skipReport = skipReport || (scriptPassed && scriptHasWarnings && !m_showWarnings);
```
jiaweig-amzn, line 444: This 2 lines can be combined into:
scriptPassed && ((!scriptHasWarnings && !m_showAll) || ( scriptHasWarnings && !m_showWarnings));

But if I understand it correctly: m_showAll includes m_showWarnings, then it can be simplified into:
!(!scriptPassed || m_showAll || (scriptHasWarnings && m_showWarnings));

Gem/Code/Source/Automation/ScriptReporter.cpp
hershey5045: Thanks for looking! I've refactored to use stringstream.

Gem/Code/Source/Automation/ScriptReporter.cpp
hershey5045: Refactored to one line

santorac (approved):

PR #377, opened 2022-03-16
Atom/mriegger/asvmsaa2x

mrieggeramzn:
Switch ASV to use 2x msaa


Gem/Code/Source/SampleComponentManager.cpp
```diff
@@ -1634,4 +1632,20 @@ namespace AtomSampleViewer
...
+        AZ_Assert(IsValidNumMSAASamples(m_numMSAASamples), "Invalid MSAA sample setting");
+        const AZ::RHI::SamplePosition Center = { 8, 8 };
+        const AZ::RHI::SamplePosition TopLeft = { 0, 0 };
+
+        AZ::RHI::MultisampleState state;
+
+        state.m_samples = static_cast<uint16_t>(m_numMSAASamples);
+        state.m_customPositionsCount = 2;
+        state.m_customPositions[0] = Center; 
+        state.m_customPositions[1] = TopLeft;
```
antonmic, line 1646: Can you fill in all 16 positions with the list I gave you? That way someone can easily switch MSAA back to a different count and all will still work.

antonmic (changes requested):

Gem/Code/Source/SampleComponentManager.cpp
mrieggeramzn: Done

Gem/Code/Source/SampleComponentManager.cpp
```diff
@@ -1634,4 +1632,31 @@ namespace AtomSampleViewer
...
     }
 
+    AZ::RHI::MultisampleState SampleComponentManager::CreateDefaultMsaaState() const
+    {
+        AZ_Assert(IsValidNumMSAASamples(m_numMSAASamples), "Invalid MSAA sample setting");
+
+        AZ::RHI::MultisampleState state;
+        state.m_samples = static_cast<uint16_t>(m_numMSAASamples);
+        state.m_customPositionsCount = m_numMSAASamples;
+        state.m_customPositions[0] = { 8, 8 };
```
moudgils, line 1642: Add a comment about the custom positions. Would be nice to see a diagram of it.

mrieggeramzn: Added a comment. Good idea since our notation is different from hlsl

moudgils: Nit. Maybe add a line saying 

"We are filling all the possible positions here even though the positions taken into consideration will be capped by m_customPositionsCount"

moudgils (approved):

moudgils:
Nit. Maybe add a line saying

"We are filling all the possible positions here even though the positions taken into consideration will be capped by m_customPositionsCount"

antonmic (approved):

PR #387, opened 2022-03-24
Jenkinsfile o3de/ASV build separation

amzn-changml:
This change separates the AR and nightly builds of ASV (or any project) between the engine and ASV. This allows the developer to see if the issue from the build is due to the engine or their change in ASV.

image

Tested in a sandbox Jenkins instance


lumberyard-employee-dm (approved):

shirangj (approved):
LGTM, just one question, what's the impact to the build time as we build o3de and ASV separately?

amzn-changml:
> LGTM, just one question, what's the impact to the build time as we build o3de and ASV separately?

Some build time delay of approx 8-10 mins (old avg time around 45 mins vs new time of 53 mins), as Jenkins needs to terminate, then reinstanciate the EC2 instances to build ASV and re-run tests. It is using the same disk for both o3de and ASV, however, so the build artifacts are preserved.

Clean build may take substantially longer, somewhere in the order of another hour, as the artifacts have to be regnerated, and built for engine and project.

brianherrera:
Just wondering if it's necessary to build the engine separately if the ASV repo is just pulling the engine code directly from the o3de repo, as opposed to using its own copy. Would the engine build pipeline here be the same result as the o3de/development pipe?

> It is using the same disk for both o3de and ASV, however, so the build artifacts are preserved.

Wondering if we can just use the snapshots from o3de/dev to cut down on duplicating the build here. However, there may be other requirements that I'm unaware of.

amzn-changml:
I'm merging this in regardless of the build status, as there are known build breakages in ASV's AR job that need to be fixed. This change only modifies the projects the AR is built on.

Pinging @jromnoa, @smurly, and @wintermute-motherbrain for awareness, however

amzn-changml:
> Just wondering if it's necessary to build the engine separately if the ASV repo is just pulling the engine code directly from the o3de repo, as opposed to using its own copy. Would the engine build pipeline here be the same result as the o3de/development pipe?

Yes it would be the same. However, the builds for ASV are a currently a combination of both the o3de and project outputs at latest. This change is trying to definitively show whether a change in o3de or a change in the project caused an overall AR failure. AR for o3de development can go red while an ASV build is occurring, and some energy is diverted to figuring out which caused the issue (however simple it may be to just look at the build alarms).

> 
> > It is using the same disk for both o3de and ASV, however, so the build artifacts are preserved.
> 
> Wondering if we can just use the snapshots from o3de/dev to cut down on duplicating the build here. However, there may be other requirements that I'm unaware of.

I believe I preserved the build snapshots for o3de dev when the incremental build script is triggered for a new volume. The bulk of the current build time is in asset processing, but not sure if it's regenerating due to a bug or if that's a known process.

brianherrera (approved):

smurly (approved):

amzn-changml:
Discussed this further with @smurly. I think we're ok to merge this in for now and handle any further improvements later.

PR #421, opened 2022-04-15
Add Eye material sample and regression test

santipaprika:
Follow-up PR for https://github.com/o3de/o3de/issues/8721 adding a new sample and test in ASV for the Eye material (added in https://github.com/o3de/o3de/pull/8129).

The sample exposes the parameters that have most influence on the eye visualization, but not all of them (which can be checked in the Material Editor/Inspector).

As can be observed, the model and textures have not been crafted by an artist, therefore these are not production-ready assets, but should be enough for the sample and test. 
Specifically, the UV layout at the back of the eyeball could be improved, but note that in most use cases this part will be hidden, so the front hemisphere would be enough.
### Sample Video Demo

https://user-images.githubusercontent.com/44426596/163544022-da9b738d-6d4e-4463-81f8-edddbe226746.mp4

  
### Test screenshot
image

Signed-off-by: Santi Paprika <santi.gonzalez@huawei.com>

### DIFF
--- a/Gem/Code/Source/EyeMaterialExampleComponent.cpp
+++ b/Gem/Code/Source/EyeMaterialExampleComponent.cpp
@@ -1,0 +1,313 @@
+/*
+ * Copyright (c) Contributors to the Open 3D Engine Project.
+ * For complete copyright and license terms please see the LICENSE at the root of this distribution.
+ *
+ * SPDX-License-Identifier: Apache-2.0 OR MIT
+ *
+ */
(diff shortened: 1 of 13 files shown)


santipaprika:
Temporarily broken until https://github.com/o3de/o3de/pull/9131 gets merged.

santorac (approved):
This looks great!

jeremyong-az (approved):

PR #423, opened 2022-04-23
Added New "ParallaxDepthArtifacts" Screenshot Tests

santorac:
Added new screenshot tests that focus on exposing depth-test artifacts in parallax surfaces.
I added a new setting to the Parallax sample that allows turning off all lights, so the surface is rendered black. This allows us to use a much tighter threshold on the screenshot comparison, since an entirely black unlit surface will not exhibit the sensitive platform differences we are used to seeing on parallax materials. This should ensure any depth-test calculation noise, which show up as gray firefly pixels, will cause the test threshold to fail.
I included many screenshots because the noise seems to be platform - and/or driver-dependent, so having more screenshots increases the chances of detecting failures.
Also made sure any IBL is cleared when closing a sample. There were some cases where the IBL from one sample could persist when opening another sample, for example when going directly from the Mesh sample to the Parallax sample.

Corresponds to these code changes https://github.com/o3de/o3de/pull/9135

The expected screenshots all look similar to this:
screenshot_1


antonmic (approved):

Scripts/ParallaxDepthArtifacts.bv.lua
```diff
@@ -0,0 +1,65 @@
...
+--
+----------------------------------------------------------------------------------------------------
+
+g_screenshotOutputFolder = ResolvePath('@user@/Scripts/Screenshots/ParallaxDepthArtifacts/')
+Print('Saving screenshots to ' .. NormalizePath(g_screenshotOutputFolder))
+
+OpenSample('Features/Parallax')
+ResizeViewport(512, 512)
+
+-- Several times there have bugs that appeared, related to inconsistent depth calculations causing unwanted clipping of pixels on parallax surfaces.
```
amzn-tommy, line 18: nit: have bugs->have been bugs

amzn-tommy: Or even better
"Several times there have been bugs that appeared," -> "There have been several bugs related..."

amzn-tommy (approved):

PR #440, opened 2022-05-05
Register the ASV script context with the script system

amzn-tommy:
My previous PR fixed the build error after the latest scripting changes, however using the ScriptSystemRequestBus::Load event to execute a script would fail because ASV's script context was not registered with the script system. This change uses the script system to create the ASV script context, so that the correct context is found when executing scripts.

Signed-off-by: Tommy Walton <waltont@amazon.com>


amzn-tommy:
Discarding this PR in favor of https://github.com/o3de/o3de-atom-sampleviewer/pull/441

PR #452, opened 2022-05-25
Antonmic/fullscreen shadows

antonmic:
ASV changes and screenshots for Fullscreen Shadow changes here: https://github.com/o3de/o3de/pull/8312


antonmic:
2022-05-25 15_44_21-AtomSampleViewer

invertednormal (approved):

Materials/DynamicMaterialTest/EmissiveMaterial.azsl
```diff
@@ -93,9 +93,6 @@ ForwardPassOutput MainPS(VSOutput IN)
     lightingData.tileIterator.Init(IN.m_position, PassSrg::m_lightListRemapped, PassSrg::m_tileLightData);
     lightingData.Init(surface.position, surface.normal, surface.roughnessLinear);
 
-    // Shadow, Occlusion
-    lightingData.shadowCoords = IN.m_shadowCoords;
```
santorac, line 97: So the new technique doesn't use shadow coords anymore? Can we remove it from the VSOutput struct?

Scripts/_FullTestSuite_.bv.lua
```diff
@@ -67,7 +68,8 @@ tests= {
     RunScriptWrapper('scripts/skinnedmesh.bv.luac'),
     RunScriptWrapper('scripts/RenderTargetTexture.bv.luac'),
     RunScriptWrapper('scripts/PassTree.bv.luac'),
-    RunScriptWrapper('scripts/ReadbackTest.bv.luac'),
+    -- Seems to cause GPU Device Lost, requires further investigation.
+    -- RunScriptWrapper('scripts/ReadbackTest.bv.luac'),
```
santorac, line 72: I don't know if commenting this out is the right approach. Can you at least confirm this was happening before your changes? If so, please open a GHI and link it in the PR description.

Scripts/_FullTestSuite_.bv.lua
```diff
@@ -57,7 +57,8 @@ tests= {
     RunScriptWrapper('scripts/streamingimagetest.bv.luac'),
     RunScriptWrapper('scripts/parallaxtest.bv.luac'),
     RunScriptWrapper('scripts/parallaxdepthartifacts.bv.luac'),
-    RunScriptWrapper('scripts/checkerboardtest.bv.luac'),
+    -- Disable checkerboard pipeline as it's quite out of date and currently a maintenance burden
+    -- RunScriptWrapper('scripts/checkerboardtest.bv.luac'),
```
santorac, line 61: Is there a plan to fix it, or to remove it entirely? It seems we need a GHI to follow up on this.

santorac (changes requested):
007_multiscatteringcompensationoff.png (and *on.png) has some new artifact at the top edge. Do you know where this is coming from? Is there some problem with the new shadow technique?
image

015_subsurfacescattering_transmission.png might also be a shadowing artifact, not sure.
image

Materials/DynamicMaterialTest/EmissiveMaterial.azsl
antonmic: Yep, when I to my optimization pass on our forward shaders I'll strip away everything that isn't needed.

Scripts/_FullTestSuite_.bv.lua
antonmic: Will discuss with the owner of this component when they return.

santorac (approved):

PR #454, opened 2022-05-26
More accurate FrameTime stats

rgba16f:
Increased the number of recorded frames to 500 to better average out ASV's variablility.
Made the average calculation use all the frame times rather than just the ones in the last 1/4 second.

### DIFF
--- a/Gem/Code/Source/SampleComponentManager.cpp
+++ b/Gem/Code/Source/SampleComponentManager.cpp
@@ -330,7 +330,6 @@ namespace AtomSampleViewer
 
     SampleComponentManager::SampleComponentManager()
         : m_imguiFrameCaptureSaver("@user@/frame_capture.xml")
-        , m_imGuiFrameTimer(FrameTimeLogSize, FrameTimeLogSize, 250.0f)
     {
         m_exampleEntity = aznew AZ::Entity();
 
@@ -486,6 +485,20 @@ namespace AtomSampleViewer
             }
             AZ_Warning("SampleComponentManager", targetSampleFound, "Failed find target sample %s", targetSampleName.c_str());
         }
+        if (commandLine->HasSwitch("timingSamples"))
+        {
+                AZStd::string timingSamplesStr = commandLine->GetSwitchValue("timingSamples", 0);
+                int timingSamplesCount = atoi(timingSamplesStr.c_str());
+                if (timingSamplesCount <= 0 || timingSamplesCount > 1000000)
+                {
+                    timingSamplesCount = static_cast<int>(FrameTimeLogSize);
+                }
+                m_imGuiFrameTimer = AZStd::make_unique<ImGuiHistogramQueue>(timingSamplesCount, timingSamplesCount, 250.0f);
+        }
+        else
+        {
+            m_imGuiFrameTimer = AZStd::make_unique<ImGuiHistogramQueue>(FrameTimeLogSize, FrameTimeLogSize, 250.0f);
+        }
 
         // Set default screenshot folder to relative path 'Screenshots'
         AZ::IO::Path screenshotFolder = "Screenshots";
@@ -543,7 +556,7 @@ namespace AtomSampleViewer
 
     void SampleComponentManager::OnTick(float deltaTime, [[maybe_unused]] AZ::ScriptTimePoint time)
     {
-        m_imGuiFrameTimer.PushValue(deltaTime * 1000.0f);
+        m_imGuiFrameTimer->PushValue(deltaTime * 1000.0f);
 
         bool screenshotRequest = false;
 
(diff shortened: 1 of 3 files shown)


invertednormal (approved):
Definitely agree with changing it to average the times of a specific number of frames instead of some number of frames over a total time. I think 500 may be excessive, but I'm ok with it for now. Would love standard deviation reporting.

santorac (approved):

rgba16f:
@invertednormal @santorac could I get a re-review since I changed this to make the number of samples configurable from the command line.

invertednormal (approved):

Gem/Code/Source/SampleComponentManager.cpp
```diff
@@ -486,6 +485,20 @@ namespace AtomSampleViewer
             }
             AZ_Warning("SampleComponentManager", targetSampleFound, "Failed find target sample %s", targetSampleName.c_str());
         }
+        if (commandLine->HasSwitch("timingSamples"))
+        {
+                AZStd::string timingSamplesStr = commandLine->GetSwitchValue("timingSamples", 0);
+                int timingSamplesCount = atoi(timingSamplesStr.c_str());
+                if (timingSamplesCount <= 0 || timingSamplesCount > 1000000)
```
moudgils, line 492: Nit. Move 1000000 to the header with a name like MaxTimingSampleCount

moudgils (approved):

Gem/Code/Source/SampleComponentManager.cpp
```diff
@@ -486,6 +485,20 @@ namespace AtomSampleViewer
             }
             AZ_Warning("SampleComponentManager", targetSampleFound, "Failed find target sample %s", targetSampleName.c_str());
         }
+        if (commandLine->HasSwitch("timingSamples"))
+        {
+                AZStd::string timingSamplesStr = commandLine->GetSwitchValue("timingSamples", 0);
+                int timingSamplesCount = atoi(timingSamplesStr.c_str());
```
gadams3, line 491: There should be more robust conversion functions than raw atoi in AZ::StringFunc.

int ToInt(const char* in);
bool LooksLikeInt(const char* in, int* pInt = nullptr);

Gem/Code/Source/SampleComponentManager.cpp
```diff
@@ -486,6 +485,20 @@ namespace AtomSampleViewer
...
+                if (timingSamplesCount <= 0 || timingSamplesCount > 1000000)
+                {
+                    timingSamplesCount = static_cast<int>(FrameTimeLogSize);
+                }
```
gadams3, line 495: You can reduce this to one line with a clamp function

Gem/Code/Source/SampleComponentManager.cpp
```diff
@@ -543,7 +556,7 @@ namespace AtomSampleViewer
 
     void SampleComponentManager::OnTick(float deltaTime, [[maybe_unused]] AZ::ScriptTimePoint time)
     {
-        m_imGuiFrameTimer.PushValue(deltaTime * 1000.0f);
+        m_imGuiFrameTimer->PushValue(deltaTime * 1000.0f);
```
gadams3, line 559: Now that this is a pointer, it should either be checked before use

Gem/Code/Source/SampleComponentManager.cpp
```diff
@@ -486,6 +485,20 @@ namespace AtomSampleViewer
...
+                int timingSamplesCount = atoi(timingSamplesStr.c_str());
+                if (timingSamplesCount <= 0 || timingSamplesCount > 1000000)
+                {
+                    timingSamplesCount = static_cast<int>(FrameTimeLogSize);
+                }
+                m_imGuiFrameTimer = AZStd::make_unique<ImGuiHistogramQueue>(timingSamplesCount, timingSamplesCount, 250.0f);
+        }
+        else
+        {
+            m_imGuiFrameTimer = AZStd::make_unique<ImGuiHistogramQueue>(FrameTimeLogSize, FrameTimeLogSize, 250.0f);
```
gadams3, line 500: Can this still be created on construction and replaced if the command line option is specified and valid?

antonmic (approved):

Gem/Code/Source/SampleComponentManager.cpp
rgba16f: Made this change

Gem/Code/Source/SampleComponentManager.cpp
rgba16f: made this change

Gem/Code/Source/SampleComponentManager.cpp
rgba16f: done

Gem/Code/Source/SampleComponentManager.cpp
rgba16f: done

Gem/Code/Source/SampleComponentManager.cpp
rgba16f: done

gadams3 (approved):

PR #459, opened 2022-06-06
Fixed a timing issue in MaterialHotReloadTest. 

santorac:
When there were many AssetBuilder.exe processes running, many reload operations seemed to happen on the same frame, which made that frame to take the entire IdleSeconds(0.25). Since the shader reloading process can span multiple frames, this was not enough of a delay. So I made the test idle for 0.25s plus 2 frames.

I also made a slight improvement to MaterialHotReloadTestComponent to check the active shader variant ID without calling GetVariant, which could have impacted the behavior of the test.

This partially addresses https://github.com/o3de/o3de-atom-sampleviewer/issues/458 

### DIFF
--- a/Gem/Code/Source/MaterialHotReloadTestComponent.cpp
+++ b/Gem/Code/Source/MaterialHotReloadTestComponent.cpp
@@ -352,7 +352,7 @@ namespace AtomSampleViewer
                 {
                     shaderVariantStatus = ShaderVariantStatus::FullyBaked;
                 }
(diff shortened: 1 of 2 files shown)


galibzon (approved):
The solution is kind of brittle. Ideally We would need an event based system that works with the ASV Lua Context. But what matters at the moment is to unblock the team. So hopefully this PR will work for everybody.

antonmic (approved):

PR #476, opened 2022-07-02
Removing unused passes

antonmic:
Companion review for https://github.com/o3de/o3de/pull/10506

Signed-off-by: antonmic <56370189+antonmic@users.noreply.github.com>


Passes/ColorblindnessSimulation.pass
```diff
@@ -1,62 +0,0 @@
-{
-    "Type": "JsonSerialization",
-    "Version": 1,
-    "ClassName": "PassAsset",
-    "ClassData": {
-        "PassTemplate": {
-            "Name": "ColorBlindPassTemplate",
-            "PassClass": "FullScreenTriangle",
```
rgba16f, line 8: I don't think we should remove the colorblindness support. Is this pass data unneeded for that support?

antonmic: This isn't colorblindness support, it's a pass to visualize how a colorblind person who see the screen. And it's not very legitimate, this was just a prototype pass made a long time ago to test out the pass system with some math that the engineer pulled off the internet for a single type of colorblindness. We can definitely add support for colorblindness, but should be done in a proper way, this was just to test the pass system more than anything else.

rgba16f (approved):

PR #478, opened 2022-07-09
Added photosensitive seizure warnings.

santorac:
A new AtomSampleComponent base class is shared by all ASV samples. If any sample needs to warn the user about its content, there is a ContentWarning variable that can be overridden. The SampleComponentManager will display the warning before opening that sample. I also added similar warnings when opening the Script Runner and Pre-Commit Wizard dialogs, because any automation script is likely to cause flashing images.

Note that I also remove the ability to use Ctrl-0 through Ctrl-9 to open samples. It only worked for the first 10 samples, and was useful in the early days of AtomSampleViewer when we couldn’t show ImGui during RHI samples. Now that all samples keep ImGui visible, we can stop maintaining this feature.

Fixes https://github.com/o3de/o3de-atom-sampleviewer/issues/398

image

## Testing

Tested entering the affected samples, clicking both OK and Cancel.
Tested opening the automation script runner and precommit wizard, clicking both OK and Cancel.
Tested the --sample command line option, clicking both OK and Cancel.
Ran the full test suite before and after the changes.
Note that running test scripts from the command line will not show any warnings, as this mode needs to run without any user interaction, and is usually only used on Jenkins.


galibzon (changes requested):
All good. Can you please change the messagebox title from "Sample Content Warning" to "Photosensitive Seizure Warning".

jeremyong-az (approved):

rgba16f (approved):

PR #496, opened 2022-08-22
Screenshots stored from nightly Jenkins ASV "Test GPU Profile" only contain Vulkan versions, not DX12 versions

jiaweig-amzn:
The original cause for the issue is screenshot tests run on dx12 and vulkan sequentially, so that the old screenshots will be overwritten by the newer ones because they have the same names.

To solve this, we can either generate different file names for screenshots, or having them stored in different folders.
The PR insert an env path in the middle of the screenshot path, specified in lua. 

Tested the full test suit, and updating local baseline images.
A single test result looks like below:
Screenshot update

### DIFF
--- a/Gem/Code/Source/Automation/ScriptManager.cpp
+++ b/Gem/Code/Source/Automation/ScriptManager.cpp
@@ -1094,7 +1094,6 @@ namespace AtomSampleViewer
         behaviorContext->Method("CaptureScreenshotWithImGui", &Script_CaptureScreenshotWithImGui);
         behaviorContext->Method("CaptureScreenshotWithPreview", &Script_CaptureScreenshotWithPreview);
         behaviorContext->Method("CapturePassAttachment", &Script_CapturePassAttachment);
-        behaviorContext->Method("QueryRenderBackend", &Script_QueryRenderBackend);
 
         // Profiling data...
         behaviorContext->Method("CapturePassTimestamp", &Script_CapturePassTimestamp);
@@ -1527,7 +1526,7 @@ namespace AtomSampleViewer
             });
     }
 
-    AZStd::string ScriptManager::Script_QueryRenderBackend()
+    AZStd::string ScriptManager::Script_GetRenderApiName()
     {
         AZ::Name renderApi = AZ::RPI::RPISystemInterface::Get()->GetRenderApiName();
         return AZStd::string(renderApi.GetCStr());
--- a/Gem/Code/Source/Automation/ScriptManager.cpp
+++ b/Gem/Code/Source/Automation/ScriptManager.cpp
@@ -1094,6 +1094,7 @@ namespace AtomSampleViewer
         behaviorContext->Method("CaptureScreenshotWithImGui", &Script_CaptureScreenshotWithImGui);
         behaviorContext->Method("CaptureScreenshotWithPreview", &Script_CaptureScreenshotWithPreview);
         behaviorContext->Method("CapturePassAttachment", &Script_CapturePassAttachment);
+        behaviorContext->Method("QueryRenderBackend", &Script_QueryRenderBackend);
 
         // Profiling data...
         behaviorContext->Method("CapturePassTimestamp", &Script_CapturePassTimestamp);
@@ -1405,9 +1406,8 @@ namespace AtomSampleViewer
         s_instance->m_scriptOperations.push(AZStd::move(operation));
     }
 
-    bool ScriptManager::PrepareForScreenCapture(const AZStd::string& path)
+    bool ScriptManager::PrepareForScreenCapture(const AZStd::string& path, const AZStd::string& suffix)
     {
-
         if (!Utils::IsFileUnderFolder(Utils::ResolvePath(path), ScreenshotPaths::GetScreenshotsFolder(true)))
         {
             // The main reason we require screenshots to be in a specific folder is to ensure we don't delete or replace some other important file.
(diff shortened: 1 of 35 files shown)


rgba16f:
ScriptAutomation has a GetRenderApiName function so similar functionality can be achieved. I think it’s better if we unify the approaches rather than have different for ASV and in game.
Why does the suffix need to be passed to c++? the file name & path is being built up in lua.

jiaweig-amzn:
> ScriptAutomation has a GetRenderApiName function so similar functionality can be achieved. I think it’s better if we unify the approaches rather than have different for ASV and in game. Why does the suffix need to be passed to c++? the file name & path is being built up in lua.

There are not only GetRenderApiName that exists in both ASV and game, but also quite a few utilities and screenshot capture functions. I think it is good to leave it for now, and we will do another PR later to unify them.
The full path is not built in lua because comparison step needs to figure out the path for the golden images. The golden images are not differentiated by test environment, there only exist one for each test. (This is because having multiple of them will become too much of a burden, and I believe such acceptable difference could be managed by the comparison threshold.) So that either we pass in the original path and the suffix, or we define a standard of how the suffix is appended. In this PR, I use "+" as an indicator for just human readability. But any standard we define could result in a conflict when extra suffix is added (say we probably have some version number info like 1.1+, then we have to change our standard).

santorac (approved):
This looks mostly okay to me, there's just a question about a possible duplicate function definition that needs to be resolved.

I don't like the fact that both lua and C++ are doing name mangling to create the paths. I'd rather lua be higher level and C++ do all of the name mangling.  But I didn't like how we were handling screenshot paths before either (my fault, I wrote it) and these changes make things even more messy IMO. But doing something cleaner would take more work and probably isn't worth the effort. This gets the job done so I say let's go with it.

Gem/Code/Source/Automation/ScriptManager.h
```diff
@@ -136,8 +136,10 @@ namespace AtomSampleViewer
         // but with "Screenshots" replaced with "ExpectedScreenshots". For example, the expected file for
         // "Scripts/Screenshots/StandardPbr/test.ppm" should be at "Scripts/ExpectedScreenshots/StandardPbr/test.ppm".
 
-        static void Script_CaptureScreenshot(const AZStd::string& filePath);
-        static void Script_CaptureScreenshotWithImGui(const AZStd::string& filePath);
+        static void Script_CaptureScreenshot(const AZStd::string& filePath, const AZStd::string& suffix);
+        static void Script_CaptureScreenshotWithImGui(const AZStd::string& filePath, const AZStd::string& suffix);
+
+        static AZStd::string Script_GetRenderApiName();
```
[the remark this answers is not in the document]
jiaweig-amzn: Fixed.

Gem/Code/Source/Automation/ScriptManager.cpp
```diff
@@ -1448,10 +1453,11 @@ namespace AtomSampleViewer
     {
         Script_SetShowImGui(false);
 
-        auto operation = [filePath]()
+        const AZStd::string& envPath = s_instance->m_envPath;
+        auto operation = [filePath, envPath]()
         {
             // Note this will pause the script until the capture is complete
-            if (PrepareForScreenCapture(filePath))
+            if (PrepareForScreenCapture(filePath, envPath))
```
rgba16f, line 1460: Wouldn't s_instance be accessible inside the lambda since it's a static variable?
envPath is a reference so there is no protection across threads reading & writing. Is there another reason to manually capture it?

rgba16f:
Can we make the expected screenshot paths api specific as well? Or is that planned for a 2nd PR?

Gem/Code/Source/Automation/ScriptManager.cpp
jiaweig-amzn: Yes, you are right. I'll update.

jiaweig-amzn:
> Can we make the expected screenshot paths api specific as well? Or is that planned for a 2nd PR?

Yes. This is to fix the issue first. We can use the same way to store expected screenshot path as env path.

rgba16f (approved):

Gem/Code/Source/Automation/ScriptManager.cpp
```diff
@@ -1433,6 +1434,11 @@ namespace AtomSampleViewer
         return true;
     }
 
+    void ScriptManager::Script_SetTestEnvPath(const AZStd::string& envPath)
+    {
+        s_instance->m_envPath = envPath;
```
santorac, line 1439: This function really ought to use an operation lambda like the other "Script_" functions do. Otherwise, there can be unexpected behavior. The envPath will be set before all other operations are run. And if a script calls SetTestEnvPath multiple times, only the last call with count, and will apply to any prior uses of CaptureScreenshot.

Scripts/AreaLightTest.bv.lua
```diff
@@ -109,7 +111,7 @@ function VaryMetallicOnly()
     SetImguiValue('AreaLightSample/Min Max Metallic', Vector2(0.0, 1.0))
 end
 
-g_screenshotOutputFolder = ResolvePath('@user@/Scripts/Screenshots/AreaLights/')
+g_screenshotOutputFolder = ResolvePath('@user@/Scripts/Screenshots/' .. g_testEnv .. '/AreaLights/')
```
santorac, line 114: We could put a utility function like this in TestEnvironment.lauc:
```
function MakeScreenshotPath(testCaseName, testStepName)
    return '@user@/Scripts/Screenshots/' .. g_testEnv .. '/' .. testCaseName .. '/' .. testStepName .. '.png'
end
```
Then remove all of the g_screenshotOutputFolder variables. And wherever CaptureScreenshot is called, do it like this:
```
CaptureScreenshot(MakeScreenshotPath("AreaLightTest", "WhateverTestStepName"))
```

santorac (commented):
I know this is already merged, but I wanted to leave a couple comments for consideration in a future PR...

PR #500, opened 2022-08-31
Move part of the screenshot process to frame capture bus so it can be shared with other projects

jiaweig-amzn:
This PR depends on
https://github.com/o3de/o3de/pull/11679

Tested ASV full test suit.
Now ASV uses the CompareScreenshot request ebus to compare images.
File paths will be managed by the frame capture system, instead of individual scripting systems, because they are too common.


jiaweig-amzn:
> 



> There is perhaps a larger discussion to be had that the paths we're using for our image capture & comparisons should be using the SettingsRegistry for their data, rather than hard coding paths.

I don't see a big difference of having the path in the setting registry or in lua.
And how can you get the runtime environment in the setreg which is part of the path?

santorac (commented):
Just a heads up, you are going to conflict with this PR https://github.com/o3de/o3de-atom-sampleviewer/pull/502

Gem/Code/Source/Automation/ScriptReporter.cpp
```diff
@@ -1168,20 +1108,29 @@ namespace AtomSampleViewer
...
+                imagesWereCompared,
+                &AZ::Render::FrameCaptureRequestBus::Events::CompareScreenshots,
                 screenshotTestInfo.m_screenshotFilePath,
-                TraceLevel::Error);
+                screenshotTestInfo.m_officialBaselineScreenshotFilePath,
+                &screenshotTestInfo.m_officialComparisonResult.m_standardDiffScore,
+                &screenshotTestInfo.m_officialComparisonResult.m_filteredDiffScore,
+                ImperceptibleDiffFilter
+            );
+            screenshotTestInfo.m_officialComparisonResult.m_finalDiffScore = screenshotTestInfo.m_officialComparisonResult.m_standardDiffScore;
```
santorac, line 1133: I think we should preserve the comment that was in the original DiffImages function:
```
// Set the final score to the standard score just in case the filtered one is ignored
```

santorac: Also, is ScriptReporter::DiffImages still used? Should it be removed now?

Scripts/AreaLightTest.bv.lua
```diff
@@ -154,7 +154,7 @@ do
         materialSetupFunction()
         
         IdleFrames(1) 
-        CaptureScreenshot(g_screenshotOutputFolder .. '/' .. lightName .. '_' .. materialName ..'.png')
+        CaptureScreenshot(g_testCaseFolder .. '/' .. lightName .. '_' .. materialName ..'.png')
```
santorac, line 157: Since you already have SetScreenshotsFolder and SetTestEnvFolder, I wonder if we should also add SetTestCaseFolder instead of making each script append it like this.

Scripts/TestEnvironment.lua
```diff
@@ -1,8 +1,11 @@
 g_renderApiName = GetRenderApiName()
 
--- surround the actual env so that it won't match other part of the path when IO is operating the file path.
-stringCollisionProtector = '___'
+g_screenshotOutputFolder = '@user@/Scripts/Screenshots/'
+g_localBaselineFolder = '@user@/Scripts/Screenshotslocalbaseline/'
+g_officialBaselineFolder = '@projectroot@/Scripts/Expectedscreenshots/'
+g_testEnv = g_renderApiName
```
santorac, line 6: Are these variables still necessary? It looks like the strings could just be passed directly to SetScreenshotFolder(), etc.

Gem/Code/Source/Automation/ScriptReporter.cpp
jiaweig-amzn: Comment added back. Unused functions removed.

Scripts/TestEnvironment.lua
jiaweig-amzn: g_screenshotOutputFolder is still used for printing out the destination folder. Others are passed directly now.

Scripts/AreaLightTest.bv.lua
jiaweig-amzn: I think inner folder structure should be managed locally. It can have multiple levels of test case, for example, it is also reasonable to add RHI/RPI as another level of test case.

Gem/Code/Source/Automation/ScriptManager.cpp
```diff
@@ -1406,36 +1409,71 @@ namespace AtomSampleViewer
...
+    }
+
     void ScriptManager::Script_SetTestEnvPath(const AZStd::string& envPath)
     {
-        s_instance->m_envPath = envPath;
+        auto operation = [envPath]()
+        {
+            AZ::Render::FrameCaptureTestRequestBus::Broadcast(
+                &AZ::Render::FrameCaptureTestRequestBus::Events::SetTestEnvPath, envPath);
+        };
```
rgba16f, line 1452: I think this creates a reference to envPath that can cause a use after free. It's probably better to duplicate the string in all the operations.

rgba16f: Actually I'm wrong, it's a copy of the referenced type.

Gem/Code/Source/Automation/ScriptManager.cpp
```diff
@@ -2057,3 +2119,5 @@ namespace AtomSampleViewer
         s_instance->m_scriptOperations.push(AZStd::move(operation));
     }
 } // namespace AtomSampleViewer
+
+#pragma optimize("", on)
```
rgba16f, line 2123: Please remove.

rgba16f (approved):
Just need to remove the one #pragma

Gem/Code/Source/Automation/ScriptManager.cpp
jiaweig-amzn: Removed here
https://github.com/o3de/o3de-atom-sampleviewer/pull/507

PR #502, opened 2022-09-06
update frame capture notification bus to use frame capture id as the address

rgba16f:
AtomSampleViewer change to match O3DE change https://github.com/o3de/o3de/pull/10592.

Testing ran ASV full suite.
DX12
image
Vulkan
image

### DIFF
--- a/Gem/Code/Source/Automation/ScriptManager.cpp
+++ b/Gem/Code/Source/Automation/ScriptManager.cpp
@@ -1428,7 +1428,6 @@ namespace AtomSampleViewer
         s_instance->m_scriptReporter.AddScreenshotTest(path, envPath);
 
         s_instance->m_isCapturePending = true;
-        s_instance->AZ::Render::FrameCaptureNotificationBus::Handler::BusConnect();
         s_instance->PauseScript();
 
         return true;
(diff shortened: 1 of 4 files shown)


Gem/Code/Source/Automation/ScriptManager.cpp
```diff
@@ -1458,11 +1457,12 @@ namespace AtomSampleViewer
             // Note this will pause the script until the capture is complete
             if (PrepareForScreenCapture(filePath, s_instance->m_envPath))
             {
-                AZ_Assert(s_instance->m_frameCaptureId == AZ::Render::FrameCaptureRequests::s_InvalidFrameCaptureId, "Attempting to start a capture while one is in progress");
-                uint32_t frameCaptureId = AZ::Render::FrameCaptureRequests::s_InvalidFrameCaptureId;
+                AZ_Assert(s_instance->m_frameCaptureId == AZ::Render::InvalidFrameCaptureId, "Attempting to start a capture while one is in progress");
+                AZ::Render::FrameCaptureId frameCaptureId = AZ::Render::InvalidFrameCaptureId;
                 AZ::Render::FrameCaptureRequestBus::BroadcastResult(frameCaptureId, &AZ::Render::FrameCaptureRequestBus::Events::CaptureScreenshot, filePath);
-                if (frameCaptureId != AZ::Render::FrameCaptureRequests::s_InvalidFrameCaptureId)
+                if (frameCaptureId != AZ::Render::InvalidFrameCaptureId)
```
santorac, line 1463: If this is false because the capture failed for some reason, we need to reset m_isCapturePending and unpause the script, to reverse what was done in PrepareForScreenCapture. Or move m_isCapturePending=true and PauseScript out of PrepareForScreenCapture.

santorac (changes requested):

Gem/Code/Source/Automation/ScriptManager.cpp
rgba16f: Updated to resume the script and log an error if the frame capture request fails here

Gem/Code/Source/Automation/ScriptManager.cpp
```diff
@@ -1635,15 +1666,10 @@ namespace AtomSampleViewer
...
     {
-         // ignore captures that are not triggered by the script manager
-        if (m_frameCaptureId == AZ::Render::FrameCaptureRequests::s_InvalidFrameCaptureId || frameCaptureId != m_frameCaptureId)
-        {
-            return;
-        }
         m_isCapturePending = false;
-        m_frameCaptureId = AZ::Render::FrameCaptureRequests::s_InvalidFrameCaptureId;
+        m_frameCaptureId = AZ::Render::InvalidFrameCaptureId;
         AZ::Render::FrameCaptureNotificationBus::Handler::BusDisconnect();
```
jiaweig-amzn, line 1673: Now it checks `frameCaptureId != AZ::Render::InvalidFrameCaptureId` before connecting to the bus, we can skip disconnecting here as well.

jiaweig-amzn (approved):

Gem/Code/Source/Automation/ScriptManager.cpp
rgba16f: It's more important to disconnect now as FrameCaptureId's are re-used. Additionally we would need further changes for the ScriptManager to be able to handle multiple active frame capture notification connections.

gadams3 (approved):

PR #540, opened 2022-10-26
Proof-of-Concept Deferred Render Pipeline

santorac:
This is a rough proof-of-concept for a deferred render pipeline, utilizing the new material pipeline abstraction system (https://github.com/o3de/sig-graphics-audio/blob/main/rfcs/MaterialPipelineAbstraction.md). The goal here is to demonstrate that the material pipeline system can allow the same .materialtype to run on multiple render pipelines, even handling the difference between Forward+ vs Deferred, and this can be done _**without additional modifications to the core engine**_.

I think eventually we'd like to clean this up and make it an even more compelling example, but for now I really need this to be checked in for testing purposes. Keeping this example working is very important for my continued development of the material pipeline system, and in fact someone inadvertently broke this deferred pipeline twice in the past couple weeks and I had to spend time tracking down the issues. By getting this committed, even in it's current rough state, we can prevent that from happening again.

- The example includes .pass files to define the deferred render pipeline as well as a .materialpipeline definition to adapt "any" material type.
- The new DeferredPipeline.materialpipeline is hooked up by overriding the registery setting O3DE/Atom/RPI/MaterialPipelineFiles.
- Updated the RPI/Mesh sample to support toggling the DeferredPipeline in addition to the LowEndPipeline. (Eventually I would like to factor this out so any sample can include similar pipeline controls).
- The MaterialPipelineTest_Basic.materialtype and MaterialPipelineTest_Animated.materialtype examples in O3DE are fully supported.
- The GBuffers are very limited at this point, only including base color, roughness, metallic, and normal. I hope to improve on this eventually, but it's enough for now.
- The overall structure of the pass and shader files are modeled after a combination of StandardPBR_ForwardPass and the DeferredFog feature.
- The DeferredLighting.pass and DeferredMaterial.pass are basisically a split of Forward.pass into two parts.
- DeferredOpaqueParent.pass is based on a copy of OpaqueParent.pass, with the material and lighting passes split up, and with subsurface scattering disable (for now) for simplicity.
- DeferredPipeline.pass is almost identical to MainPipeline.pass, it only has a different name and uses DeferredOpaqueParentTemplate instead of OpaqueParentTemplate.
- Added a new screenshot regression test that uses the deferred pipeline, to make sure it doesn't get broken again.
- There is a watermark in the deferred lighting pass, otherwise it's impossible to tell whether the deferred pipeline is actually being used. We need a way to disable this watermark by default and only enable it for testing, but I want to get this regression test committed ASAP.


Passes/PrototypeDeferredPipeline/DeferredLighting.azsl
```diff
@@ -0,0 +1,160 @@
...
+    // Apply Image Based Lighting (IBL)
+    // TODO(DeferredPOC): how can we support a localized reflection probe in deferred?
+    ReflectionProbeData unusedReflectionProbe; 
+    TextureCube unusedReflectionProbeCubeMap;
+    ApplyIBL(surface, lightingData, true, false, unusedReflectionProbe, unusedReflectionProbeCubeMap);
+
+    // Finalize Lighting
+    lightingData.FinalizeLighting();
+
+    PbrLightingOutput lightingOutput = GetPbrLightingOutput(surface, lightingData, alpha);
```
antonmic, line 147: Can just call EvaluateLighting_StandardPBR, no need to duplicate the code, or does this do something different?

Passes/PrototypeDeferredPipeline/DeferredLighting.azsl
```diff
@@ -0,0 +1,160 @@
...
+
+    // TODO(DeferredPOC): We don't actually have the vertex normal, so just use the per-pixel normal for now
+    surface.vertexNormal = normal;
+    surface.normal = normal;
+    surface.roughnessLinear = roughnessMetallic.x;
+    float specularF0Factor = 0.5f; // TODO(DeferredPOC): Consider passing the actual specF0 through a GBuffer
+    surface.SetAlbedoAndSpecularF0(baseColor, specularF0Factor, roughnessMetallic.y);
+    surface.clearCoat.InitializeToZero();
+        
+    surface.CalculateRoughnessA();
```
antonmic, line 113: Should be it's own function that returns Surface
Eventually we'll want to have that function in it's own file so we can change the GBuffer layout in a single location without touching other files

Materials/Pipelines/PrototypeDeferredPipeline/DeferredMaterialPass.azsli
```diff
@@ -0,0 +1,103 @@
...
+    // These are the values we expect MaterialFunction_EvaluateSurface to potentially replace.
+    surface.normal = surface.vertexNormal;
+    surface.roughnessLinear = 0.0;
+    float3 baseColor = float3(0.5, 0.5, 0.5);
+    float metallic = 0.0;
+    float specularF0Factor = 0.5f;
+    surface.SetAlbedoAndSpecularF0(baseColor, specularF0Factor, metallic);
+    surface.clearCoat.InitializeToZero();
+
+    MaterialFunction_AdjustSurface(surface);
```
antonmic, line 88: Same comment as discussed over slack, isn't MaterialFunction_AdjustSurface just re-creating the functionality of EvaluateSurface_ ? Don't really understand the need for a MaterialFunction_AdjustSurface function, maybe I'm missing something

antonmic (approved):

Passes/PrototypeDeferredPipeline/DeferredMaterial.pass
```diff
@@ -0,0 +1,144 @@
...
+                            "Pass": "Parent",
+                            "Attachment": "PipelineOutput"
+                        }
+                    },
+                    "MultisampleSource": {
+                        "Pass": "This",
+                        "Attachment": "DepthStencilInputOutput"
+                    },
+                    "ImageDescriptor": {
+                        "Format": "R16G16B16A16_FLOAT",
```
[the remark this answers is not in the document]
santorac: Okay, thanks, I'll look into this in an upcoming iteration. For now I'm being really loose about how I define the g-buffers since it is just a proof of concept used for testing.

Passes/PrototypeDeferredPipeline/DeferredLighting.azsl
santorac: Yes, I believe I made this file before the big shader refactor, and I can come back and refactor this code later.

Passes/PrototypeDeferredPipeline/DeferredLighting.azsl
santorac: Yes, I believe I made this file before the big shader refactor, and I can come back and refactor this code later.

Materials/Pipelines/PrototypeDeferredPipeline/DeferredMaterialPass.azsli
santorac: Yes, I recognize it looks a little awkward right now. It's because the EvaluateSurface_ functions and MaterialFunction_AdjustSurface functions are part of two different refactoring efforts. The paradigms of those efforts are a bit different but also have a lot of overlap. So for now, these functions are serving similar purposes with one just wrapping another. I still have a lot of iteration to do, and in the end I expect we will merge the two paradigms together, for a nice cohesive whole, and then there will be only one of these surface evaluation functions.

Gem/Code/Source/MeshExampleComponent.cpp
```diff
@@ -98,24 +98,68 @@ namespace AtomSampleViewer
...
-        m_scene->RemoveRenderPipeline(m_originalPipeline->GetId());
+        m_lowEndPipeline->SetDefaultView(prevPipeline->GetDefaultView());
+        m_scene->RemoveRenderPipeline(prevPipeline->GetId());
 
+        m_imguiScope = {}; // I'm not sure why this is needed. Something must be wrong in the ImGuiActiveContextScope class
         m_imguiScope = AZ::Render::ImGuiActiveContextScope::FromPass({ m_lowEndPipeline->GetId().GetCStr(), "ImGuiPass" });
     }
 
-    void MeshExampleComponent::DeactivateLowEndPipeline()
+    void MeshExampleComponent::ActivateDeferredPipeline()
```
jeremyong-az, line 136: Suggest an `ActivatePipeline` that accepts an argument to the pipeline to activate

Passes/PrototypeDeferredPipeline/DeferredLighting.azsl
```diff
@@ -0,0 +1,160 @@
...
+    float zDepth = PassSrg::m_depthStencilTexture.Load(screenCoords, sampleIndex).r;
+    float3 surfacePosWS = WorldPositionFromDepthBufferMS(PassSrg::m_depthStencilTexture, zDepth, IN.m_position.xy).xyz;
+    float3 encodedNormal = PassSrg::m_normal.Load(screenCoords, sampleIndex).xyz;
+    float3 normal = DecodeNormalSignedOctahedron(encodedNormal.xyz);
+    float2 roughnessMetallic = PassSrg::m_roughnessMetal.Load(screenCoords, sampleIndex).rg;
+    float3 baseColor = PassSrg::m_baseColor.Load(screenCoords, sampleIndex).rgb;
+    
+    if(o_deferredPipelineWatermark)
+    {
+        if(abs(IN.m_texCoord.x-0.5) < 0.01 || abs(IN.m_texCoord.y-0.5) < 0.01)
```
jeremyong-az, line 91: Suggest `any(abs(IN.m_texCoord - 0.5f) < 0.01)` (also, fun viz btw)

jeremyong-az (approved):

jeremyong-az:
Looks really clean considering how it interacts with the material system, will make it pretty easy to iterate on the gbuffer layout later also

Gem/Code/Source/MeshExampleComponent.cpp
santorac: At some point I plan to move these controls to a sharable location, where other Samples can include the same functionality. I'll see if I can clean up the controls at that time.

PR #550, opened 2022-11-16
Remove material hot reload tasks along with material self reloading and notification buses

gadams3:
This change is removing the material hot reload test from the sample viewer corresponding to the buses and support being removed

Signed-off-by: gadams3 <guthadam@amazon.com>

related changes found here https://github.com/o3de/o3de/pull/13186


cgalvan (approved):

jeremyong-az (approved):

rgba16f (approved):

santorac:
Hot reload is relatively fragile and these tests have been gradually expanded for quite a long time to cover some very specific nuances and edge cases. Obviously some of this is no longer possible to test in AtomSampleViewer since material reloading is now handled at the component application level. Are there plans to add automated hot reload tests elsewhere, in the AutomatedTesting project?

Also, this test didn't just cover material reloading, it also covered shader reloading, which _is_ still handled by the Material class. Maybe you are planning to remove shader reload handling from the Material class as well?

PR #556, opened 2022-12-02
Merge openxr branch to development

amzn-phist:
This includes all the work that went into the ASV project for XR.  We kept everything in a separate branch, but this includes a final piece that allows us to merge it back to development.

There's two samples here:
1) RPI Sample (`-rhi=vulkan -openxr=enable -sample=Features/OpenXR`)
2) RHI Sample (`-rhi=vulkan -openxr=enable -sample=RHI/OpenXr`)

It is advised that anyone working on these samples must first get and register the two XR Gems from [o3de-extras](https://github.com/o3de/o3de-extras) repo.  Once these external XR gems are registered to your manifest, CMake will pick them up for this project.


moudgils (approved):

Gem/Code/Source/SampleComponentManager.cpp
```diff
@@ -339,6 +342,7 @@ namespace AtomSampleViewer
             NewFeaturesSample<SkinnedMeshExampleComponent>("SkinnedMesh"),
             NewFeaturesSample<SsaoExampleComponent>("SSAO"),
             NewFeaturesSample<SSRExampleComponent>("SSR"),
+            NewFeaturesSample<XRRPIExampleComponent>("OpenXR"),
```
moudgils, line 345: Recommend making this change. It will grey out the sample if the xrsystem is not registered. Recommend the same for the XRExampleComponent above.
`NewRHISample<XRRPIExampleComponent>("OpenXR", [] { return AZ::RPI::RPISystemInterface::Get()->GetXRSystem()!=nullptr; }),`

moraaar (approved):

Gem/Code/Source/SampleComponentManager.cpp
amzn-phist: Okay, I'll get these done and post an update.

PR #559, opened 2022-12-07
Removes an empty static lib CMake target

amzn-phist:
Removes the `AtomSampleViewer.Lib.Static` CMake target

- Source files were removed from the lib, but that caused build errors: `>LINK : fatal error LNK1181: cannot open input file '..\..\..\..\lib\profile\AtomSampleViewer.Lib.Static.lib'`
- Removing the target from CMake fixes the issue.


santorac:
Here's the offending PR: https://github.com/o3de/o3de-atom-sampleviewer/pull/557

santorac (approved):

junbo75 (approved):

PR #565, opened 2022-12-20
Small fixes with RPI/StreamingImage sample 

VickyAtAZ:
Fixed srg compile warnings. 
Removed the usage of deleted variables in HeapMemoryUsage.


gadams3 (approved):

Gem/Code/Source/StreamingImageExampleComponent.cpp
```diff
@@ -457,8 +457,15 @@ namespace AtomSampleViewer
             if (m_3dImages.size() == 0)
             {
                 Create3dimages();
+                if (m_3dImages.size() == 0)
```
moudgils, line 460: we already have if (m_3dImages.size() == 0) above so no need for it here.

moudgils (approved):

jiaweig-amzn (approved):

Gem/Code/Source/StreamingImageExampleComponent.cpp
VickyAtAZ: The image may be created failed. And that’s the second condition for.

PR #568, opened 2022-12-21
Add support of Variable Rate Shading image on the Features/OpenXR sample

akioCL:
The MultiViewForward.pass now supports variable rate shading attachment. Added support to SampleComponentManager to build the VRS texture and use in the Pipeline for XR.

o3de changes [here](https://github.com/o3de/o3de/pull/13811)
XR gem change [here](https://github.com/o3de/o3de-extras/pull/79)

Signed-off-by: Akio Gaule <10719597+akioCL@users.noreply.github.com>


Gem/Code/Source/SampleComponentManager.cpp
```diff
@@ -1749,18 +1752,87 @@ namespace AtomSampleViewer
...
+                        RHI::XRFoveatedLevel foveatedType = RHI::XRFoveatedLevel::None;
+                        // Check settings registry for the foveated level
+                        if (AZ::SettingsRegistryInterface* settingsRegistry = AZ::SettingsRegistry::Get())
+                        {
+                            if (u64 foveatedLevel; settingsRegistry->Get(foveatedLevel, RHI::XRFoveatedLevelKey))
+                            {
+                                AZ_Assert(
+                                    foveatedLevel <= static_cast<u64>(RHI::XRFoveatedLevel::High),
+                                    "Invalid foveated level %d", foveatedLevel);
+                                foveatedType = static_cast<RHI::XRFoveatedLevel>(foveatedLevel);
```
moraaar, line 1815: To print `foveatedLevel` which is a AZ::u64 (uint64_t):

````
"Invalid foveated level %" PRIu64, foveatedLevel);
````

**Review by moraaar — APPROVED** [review]

File: Registry/openxr.setreg [hunk 2]
```diff
@@ -0,0 +1,9 @@
+{
+    "O3DE": {
+        "Atom": {
+            "RHI": {
+                "XRFoveatedLevel" : 2
+            }
+        }
+    }
```
moraaar, line 8: we were adding all the XR parameters under `O3DE/Atom/OpenXR`. I'll comment it in the other [PR](https://github.com/o3de/o3de/pull/13811) where this is added.

Gem/Code/Source/SampleComponentManager.cpp
```diff
@@ -1749,18 +1752,87 @@ namespace AtomSampleViewer
...
+            if (RHI::CheckBitsAll(device->GetFeatures().m_shadingRateTypeMask, RHI::ShadingRateTypeFlags::PerRegion) && !m_vrsTexture)
+            {
+                // Need to fill the contents of the Variable shade rating image.
+                const AZStd::shared_ptr<const RPI::PassTemplate> forwardTemplate =
+                    RPI::PassSystemInterface::Get()->GetPassTemplate(Name("MultiViewForwardPassTemplate"));
+                // Find the Shading Rate Attachment
+                Data::Asset<RPI::AttachmentImageAsset> vrsImageAsset;
+                for (const auto& imageAttachment : forwardTemplate->m_imageAttachments)
+                {
+                    if (RHI::CheckBitsAll(imageAttachment.m_imageDescriptor.m_bindFlags, RHI::ImageBindFlags::ShadingRate))
```
rgba16f, line 1765: This looks to be a duplicate of the code in the BootstrapSystemComponent, that seems like a reason the default VRS setup should be moved into Atom proper rather than being passed in from the outside (LyIntegration, or ASV).

rgba16f (approved):

Gem/Code/Source/SampleComponentManager.cpp
```diff
@@ -1749,18 +1752,87 @@ namespace AtomSampleViewer
...
+            }
+
             RPI::RenderPipelineDescriptor xrPipelineDesc;
             xrPipelineDesc.m_mainViewTagName = "MainCamera";
             xrPipelineDesc.m_renderSettings.m_multisampleState.m_samples = static_cast<uint16_t>(m_numMSAASamples);
 
             // Build the pipeline for left eye
             xrPipelineDesc.m_name = "RPISamplePipelineXRLeft";
-            xrPipelineDesc.m_rootPassTemplate = "LowEndPipelineXRLeftTemplate";
+            xrPipelineDesc.m_rootPassTemplate = "MultiViewXRLeftPipelineTemplate";
```
moudgils, line 1830: Can you delete LowEndPipelineXRLeft.pass and LowEndPipelineXRRight.pass as well? No one should be using them anymore.

Gem/Code/Source/SampleComponentManager.cpp
```diff
@@ -1749,18 +1752,87 @@ namespace AtomSampleViewer
...
         {
+            RHI::Ptr<RHI::Device> device = Utils::GetRHIDevice();
+            if (RHI::CheckBitsAll(device->GetFeatures().m_shadingRateTypeMask, RHI::ShadingRateTypeFlags::PerRegion) && !m_vrsTexture)
+            {
+                // Need to fill the contents of the Variable shade rating image.
+                const AZStd::shared_ptr<const RPI::PassTemplate> forwardTemplate =
+                    RPI::PassSystemInterface::Get()->GetPassTemplate(Name("MultiViewForwardPassTemplate"));
+                // Find the Shading Rate Attachment
+                Data::Asset<RPI::AttachmentImageAsset> vrsImageAsset;
+                for (const auto& imageAttachment : forwardTemplate->m_imageAttachments)
```
moudgils, line 1763: Maybe not something we need to solve now but what happens if there are multiple passes with different shading rate attachments? At the moment only the first one will get updated.

moudgils (approved):

PR #571, opened 2023-01-12
Update screenshots for material pipeline

jiaweig-amzn:

There doesn't seem to be any rendering issues for those screenshots (see below). Some of those under threshold are also close to the boundary. So update them all.
If the tests fail again, we may consider increase the tolerance level.

pipeline_diff


rgba16f (approved):

rgba16f (approved):

amzn-tommy (approved):

santorac:
I would just update the threshold. If it was failing on your computer and you just update the screenshots, then it's going to fail on the computer of whoever captured the original screenshots.

jiaweig-amzn:
> I would just update the threshold. If it was failing on your computer and you just update the screenshots, then it's going to fail on the computer of whoever captured the original screenshots.

Ok. I'll increase the threshold by 1 level as well.

PR #581, opened 2023-02-07
Change screenshot APIs

jiaweig-amzn:

Base changes from:
https://github.com/o3de/o3de/pull/14444

Also removed `m_standardDiffScore` and `m_filteredDiffScore` from comparison result, and use a single `m_diffScore`. It will pick up which score to use right after calling the comparison Ebus.

### DIFF
--- a/Gem/Code/Source/Automation/ScriptManager.cpp
+++ b/Gem/Code/Source/Automation/ScriptManager.cpp
@@ -690,7 +690,7 @@ namespace AtomSampleViewer
         ImGui::Text("Screenshot Name: %s", screenshotTest.m_officialBaselineScreenshotFilePath.c_str());
 
         ImGui::Separator();
-        ImGui::Text("Diff Score: %f", screenshotTest.m_officialComparisonResult.m_finalDiffScore);
+        ImGui::Text("Diff Score: %f", screenshotTest.m_officialComparisonResult.m_diffScore);
 
         // TODO: Render screenshots in ImGui.
         ImGui::Separator();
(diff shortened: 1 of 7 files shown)


Gem/Code/Source/Automation/ScriptManager.cpp
```diff
@@ -1544,27 +1565,39 @@ namespace AtomSampleViewer
...
-                    AZ_Error("Automation", false, "Failed to initiate frame capture for '%s'", screenshotFilePath.c_str());
+                    ReportScriptError(AZStd::string::format("Failed to initiate frame capture for '%s'", screenshotFilePath.c_str()));
                     s_instance->m_isCapturePending = false;
                     s_instance->m_frameCaptureId = AZ::Render::InvalidFrameCaptureId;
                     s_instance->ResumeScript();
+                    return;
                 }
+
+                s_instance->m_frameCaptureId = capOutcome.GetValue();
+                s_instance->AZ::Render::FrameCaptureNotificationBus::Handler::BusConnect(s_instance->m_frameCaptureId);
```
invertednormal, line 1600: This code is identical to the code above. Should be able to put in a function.

Gem/Code/Source/Automation/ScriptManager.cpp
```diff
@@ -1588,27 +1621,39 @@ namespace AtomSampleViewer
...
```
invertednormal, line 1656: Again, nearly identical to above code, only the ebus call differs.

Gem/Code/Source/Automation/ScriptManager.cpp
```diff
@@ -1699,27 +1744,40 @@ namespace AtomSampleViewer
...
```
invertednormal, line 1779: Same as above.

invertednormal (changes requested):
Looks good other than some duplicated code which should be pretty easy to factor out.

Gem/Code/Source/Automation/ScriptManager.cpp
jiaweig-amzn: Thanks for pointing this out. My next task is to merge ScriptManager into ScriptAutomation Gem, so that this part of the code will be reorganized. I'd skip refactoring in this PR and do a deeper one in the next PR.

invertednormal (approved):
Talked offline with @jiaweig-amzn and he'd like to address the comments in a separate PR with some other cleanup work. Approving since there's no real functionality issues with the code as-is.

Gem/Code/Source/Automation/ScriptManager.cpp
```diff
@@ -1413,12 +1413,21 @@ namespace AtomSampleViewer
...
             &AZ::Render::FrameCaptureTestRequestBus::Events::BuildScreenshotFilePath,
             imageName, true);
 
+        if (!pathOutcome.IsSuccess())
+        {
+            ReportScriptError(pathOutcome.GetError().m_errorMessage);
+            return false;
+        }
+
+        AZStd::string fullFilePath = pathOutcome.IsSuccess() ? pathOutcome.GetValue() : "";
```
rgba16f, line 1429: nit: no real need for the ternary operator, pathOutcome has already been checked for success so it can be assumed.

Can wait for any further work.

rgba16f (approved):

PR #584, opened 2023-02-09
Added test result diff support on Linux

galibzon:
with Beyond Compare (aka bcompare)

bcompare is already assumed to be installed
on MacOS and Windows.

In the future we can make a PySide based
tool that leverages applications like
idiff: https://manpages.ubuntu.com/manpages/jammy/en/man1/idiff.1.html and have a fully open source alternative for all
major Host OSes supported by O3DE.

Signed-off-by: galibzon <66021303+galibzon@users.noreply.github.com>

### DIFF
--- a/Gem/Code/Source/Platform/Linux/Utils_Linux.cpp
+++ b/Gem/Code/Source/Platform/Linux/Utils_Linux.cpp
@@ -7,6 +7,8 @@
  */
 #include <Utils/Utils.h>
 
+#include <sys/types.h>
+#include <sys/wait.h>
 #include <AzCore/PlatformIncl.h>
 
 namespace AtomSampleViewer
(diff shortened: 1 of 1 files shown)


moudgils (approved):

Gem/Code/Source/Platform/Linux/Utils_Linux.cpp
```diff
@@ -20,7 +22,43 @@ namespace AtomSampleViewer
 
         bool RunDiffTool(const AZStd::string& filePathA, const AZStd::string& filePathB)
         {
-            return false;
+            bool result = true;
+            AZStd::string executablePath = "/usr/bin/bcompare";
+
+            // Fork a process to run Beyond Compare app
+            pid_t childPid = fork();
```
rgba16f, line 29: why is forking necessary to launch another app?

rgba16f (approved):

jiaweig-amzn (approved):

Gem/Code/Source/Platform/Linux/Utils_Linux.cpp
galibzon: That's the UNIX way. Pretty standard stuff, and it's the same code as in MacOS.

PR #597, opened 2023-03-17
Turning off msaa by default in AtomSampleViewer

invertednormal:
This change sets the default number of msaa samples in asv to 1, thereby turning it off for any platforms or samples that don't specify otherwise. Removing 'GetDefaultNumMSAASamples()' from `SampleComponentManager`, platforms and samples can instead call `SetDefaultNumMSAASamples()` if they want an msaa sample count other than one, making it so platforms don't have to implement the abstract function.

Updating all screenshots that were previously passing and were affected by the msaa change. All failing screenshots have been left but will need to be updated when their original issue is resolved.

A few special cases:
- PassTest was testing attachments on several passes that were msaa specific. These have been changed to still test the various formats from before, but on passes more likely to be in every pipeline. This resulted in the removal of depthStencilMs.png, depthStencilResolve.png, and specularResolved.png
- The MultiPipeline sample was explicitly setting the msaa count of it's second pipeline to 4, it now gets that information from the SampleComponentManager instead.
- There were directional shadow issues on the Shadowed Sponza test that have disappeared. This is likely because fullscreen shadows assume default sample locations, and asv used to have custom sample locations.

Here's a list of all the screenshots that were _not_ updated
- checkerboard
- culling and lod
- dynamic material (never finishes)
- exposure
- material:
  - materialpipelinetest_enhanced_tintedtransparent_lowendpipeline
  - materialpipelinetest_standard_deferredpipeline
  - materialpipelinetest_basic_deferredpipeline
  - materialpipelinetest_enhanced_deferredpipeline
  - materialpipelinetest_enhanced_tintedtransparent_lowendpipeline
  - materialpipelinetest_standard_pdo_deferredpipeline
  - materialpipelinetest_enhanced_pdo_deferredpipeline
  - materialpipelinetest_standard_tintedtransparent_lowendpipeline
  - reflectionprobevisualization
  - 002_wrinkle_regression_test
  - 015_subsurfacescattering_transmission_thin
  - 015_subsurfacescattering_transmission
  - 015_subsurfacescattering
  - 007_multiscatteringcompensationoff
  - 007_multiscatteringcompensationon
  - minimalpbr_reddielectric
  - 001_defaultwhite
  - minimalpbr_default
- multiscene : Start_MultiScene2
- scenereloadsoaktest (never finishes)
- shaderreloadsoaktest (crashes)
- shadowteset (failures seem to be a race condition, one of the shadows doesn't appear in the test but the sample itself is fine)
  - spot_filter_method
  - spot_shadowmap_size
- skinnedmeshmotionvector (never finishes)


moudgils (approved):

Gem/Code/Source/SampleComponentManager.cpp
```diff
@@ -1744,7 +1757,7 @@ namespace AtomSampleViewer
             pipelineDesc.m_materialPipelineTag = GetMaterialPipelineName();
             pipelineDesc.m_mainViewTagName = "MainCamera";
             pipelineDesc.m_allowModification = true;
-            pipelineDesc.m_renderSettings.m_multisampleState.m_samples = static_cast<uint16_t>(m_numMSAASamples);
+            pipelineDesc.m_renderSettings.m_multisampleState.m_samples = static_cast<uint16_t>(m_numMsaaSamples);
```
VickyAtAZ, line 1760: nit: this static_cast<uint16_t> can be removed.

Gem/Code/Source/SampleComponentManager.cpp
```diff
@@ -1773,7 +1786,7 @@ namespace AtomSampleViewer
 
             RPI::RenderPipelineDescriptor xrPipelineDesc;
             xrPipelineDesc.m_mainViewTagName = "MainCamera";
-            xrPipelineDesc.m_renderSettings.m_multisampleState.m_samples = static_cast<uint16_t>(m_numMSAASamples);
+            xrPipelineDesc.m_renderSettings.m_multisampleState.m_samples = static_cast<uint16_t>(m_numMsaaSamples);
```
VickyAtAZ, line 1789: same as above

VickyAtAZ (approved):
Few small comments.

PR #616, opened 2023-04-08
Restore Standalone target

AMZN-alexpete:
The AtomSampleViewerStandalone target was removed in a previous change,  this adds it back.
Fixes https://github.com/o3de/o3de-atom-sampleviewer/issues/600

Tested by configuring and building the target, and verifying it appeared in Visual Studio

### DIFF
--- a/project.json
+++ b/project.json
@@ -14,7 +14,8 @@
     "icon_path": "preview.png",
     "engine": "o3de",
     "external_subdirectories": [
-        "Gem"
+        "Gem",
+        "Standalone"
     ],
     "gem_names": [
         "Maestro",


moudgils (approved):

moudgils:
This may need to be moved to dev branch manually as ASV is not covered by frequent integrations

rgba16f:
I don't think we should bring this back. AtomSampleViewer.GameLauncher provides the same functionality and requires no extra maintenance from the Atom team.

jromnoa:
The python tests cannot run without the standalone version, the `--runtestsuite` and `--exitontestend` CLI args that it depends on are not supported by `AtomSampleViewer.GameLauncher.exe`

rgba16f:
@AMZN-alexpete should we re-target this at Development? It seems like it's missed the window for Stabilization.

PR #629, opened 2023-06-15
Making changes to allow gems to create and register sample components in ASV

antonmic:
Made changes to allow gems to create and register their own samples. This required the following changes:
1. Remove static ScriptManager and static ScriptableImGui instances from their respective .cpp as these static global variables were not playing nicely across gems. Replaced them instead with GetInstance() functions, which has a static pointer that it initializes through a bus call if it's not already set. 
2. Created a new SampleComponentSingletonRequests bus to initialize the static pointers above through the SampleComponentManager.
3. In this new bus, added a function for registering sample components so that gems can register their samples with the SampleComponentManager. This function is in this new bus and not in the SampleComponentManagerRequestBus because that bus does not get hooked up during the activate function but later during the SampleComponentManager::ActivateInternal() function, which doesn't guarantee order dependency with gem system components that depend on the SampleComponentManagerService

Test: can successfully create new ASV sample using custom gem.

### DIFF
--- a/Gem/Code/Source/Automation/ScriptManager.cpp
+++ b/Gem/Code/Source/Automation/ScriptManager.cpp
@@ -39,20 +39,23 @@
 
 namespace AtomSampleViewer
 {
-    ScriptManager* ScriptManager::s_instance = nullptr;
-
     ScriptManager::ScriptManager()
         : m_scriptBrowser("@user@/lua_script_browser.xml")
     {
     }
 
+    ScriptManager* ScriptManager::GetInstance()
+    {
+        static ScriptManager* s_instance = nullptr;
+        if (!s_instance)
+        {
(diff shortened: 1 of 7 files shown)


AMZN-alexpete (approved):
Nice!
Not necessary for this PR but would be nice to fix the spelling and change `AZ_DEBUG_STATIC_MEMEBER` to `AZ_DEBUG_STATIC_MEMBER` if it's still used.

antonmic:
> Nice! Not necessary for this PR but would be nice to fix the spelling and change `AZ_DEBUG_STATIC_MEMEBER` to `AZ_DEBUG_STATIC_MEMBER` if it's still used.
> 
> Another option would be to use the `AZ::Interface` class which is designed to allow singletons across modules: https://github.com/o3de/o3de/blob/development/Code/Framework/AzCore/AzCore/Interface/Interface.h

Thanks for the review!
AZ::Interface is a great callout, however in these two cases the static members where used only within static functions of the classes themselves, with one already being created in the SampleComponentManager

galibzon (approved):

VickyAtAZ (approved):

PR #659, opened 2024-01-27
Ensure directional light shadow is turned on by default in the ASV examples

### LINKED ISSUE #657: Many tests have screenshot faliures due to missing directional light shadow
Run ASV __full_test_suites__. Notice there are good amount of screenshot failures due to missing directional light shadow. 

To repro the issue,  run shadowtest.lua. 

image

It may relate to these two commits:
6404925e1bf82fa2eb919ebf943b7d964d7e2dae	12/7/2023 6:03:08 AM +00:00	Add support for lighting channels
298cb5945b35fdc2d016501b5d16235536332292	11/6/2023 8:11:37 AM +00:00	Support controlling which directional light to cast shadow (#16913)

hosea1008:
This should fix some of the failed test cases in #657 

Partially Fix #657 

This PR is to set the directional light shadow to be enabled by default due to [https://github.com/o3de/o3de/pull/16913](https://github.com/o3de/o3de/pull/16913) grants access to control whether or not a directional light is casting shadow.

image

### DIFF
--- a/Gem/Code/Source/CullingAndLodExampleComponent.cpp
+++ b/Gem/Code/Source/CullingAndLodExampleComponent.cpp
@@ -224,6 +224,7 @@ namespace AtomSampleViewer
             dirLightFP->SetDirection(handle, lightTransform.GetBasis(1));
 
             dirLightFP->SetRgbIntensity(handle, Render::PhotometricColor<Render::PhotometricUnit::Lux>(m_directionalLightIntensity * DirectionalLightColor));
+            dirLightFP->SetShadowEnabled(handle, m_dirShadowEnabled);
             dirLightFP->SetCascadeCount(handle, s_cascadesCountDefault);
             dirLightFP->SetShadowmapSize(handle, s_shadowmapSizes[s_shadowmapSizeIndexDefault]);
             dirLightFP->SetDebugFlags(handle,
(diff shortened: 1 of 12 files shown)


moudgils (approved):

galibzon (approved):

VickyAtAZ:
I wonder whether it makes sense to enable shadow by default within the engine. What was the reason to disable the shadow by default?

VickyAtAZ (approved):

hosea1008:
> I wonder whether it makes sense to enable shadow by default within the engine. What was the reason to disable the shadow by default?

Hi, I am not sure you are wondering about enabling or disabling, let me explain a bit:

Enable directional shadow in the engine: The engine currently only support one directional light shadow, but it allows us to add multiple directional lights, if the artist add more than one dir light, when entering game mode, an arbitrary one may cast shadow, this is not controllable, so I was adding the switch to turn on/off the shadow to let the artist decide which one dir light to cast shadow, and the dir light shadow is enabled by default in the engine.

Disable the dir light shadow in ASV: This is not an intended behavior, it seems that the ASV is accessing the dir light fp directly, skipping the configuration and setting part, and the dir light shadow is enabling in the configuration. This commit is to fix this missing config.

hosea1008:
Hi, I am wondering how this PR is going to be merged after collecting enough reviews and all checks passed? This PR would fix some of the known issues as discussed above, I am worrying it would be conflicted after waiting for long time.

